PR #142, opened 2020-07-19
Update javascript-quiz.md

abmaxwell:
Q.31 is incorrect. Per https://www.w3schools.com/js/js_reserved.asp "Catch", "Function" and "This" are reserved words while "array" is not.


abmaxwell:
Corrected Q.31 per https://www.w3schools.com/js/js_reserved.asp

PR #177, opened 2020-07-27
Update css-quiz

tainenko:
Add 5 css-quiz


tainenko:
Add css and bash questions.

PR #184, opened 2020-07-29
- new questions added

adriadejuan:


JustinStolle:
You have the wrong answer for Q22. The underscore will match any single character so you must bracket it to match a literal `_`, otherwise you could potentially return a serial number ending in '1013', for example.

Ebazhanov:
Good catch @JustinStolle thank you! What would be your choice?

```
#### Q22. You need to write a query that returns all products that have a SerialNumber ending with "10_3". Which WHERE clause should you use to fill in the blank in this query?

    SELECT ProductID, ProductName, SerialNumber
    FROM Products______ ;

- [?] `WHERE SerialNumer LIKE '%10_3'`
- [?] `WHERE SerialNumer LIKE ('%10'+'_'+'3')`
- [?] `WHERE SerialNumer LIKE '%10"_"3'`
- [?] `WHERE SerialNumer LIKE '%10[_]3'`
```

JustinStolle:
It would be the only answer that puts the underscore in brackets, as in `[_]`. See [Using Wildcard Characters As Literals](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15#using-wildcard-characters-as-literals) for a written explanation.

Ebazhanov:
sorry, still not sure that I understand you correctly, I added your explanation https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/master/t-sql/t-sql-quiz.md
what else we can improve? please feel free to create PR or write some more words here )))

PR #281, opened 2020-08-15
Update rest-api-quiz.md

LarryChiem:
#### Q18. What is one benefit that OAuth provides over an API key approach?
- A token is scoped to the use case. <--


Ebazhanov:
> #### Q18. What is one benefit that OAuth provides over an API key approach?
> * A token is scoped to the use case. <--

@all-contributors please add @LarryChiem for design

PR #305, opened 2020-08-19
Update JavaScript Quiz

flyingSquirrel-dev:
I am not sure that the quiz no.50 has any problems to run.
Code looks good, and JavaScript Engine is able to understand it I guess.

However, I think The condition in the ternary is using the assignment operator. is close to the correct answer because the others can not be the answer.

Please give me any feedbacks! :)

### DIFF
--- a/javascript/javascript-quiz.md
+++ b/javascript/javascript-quiz.md
@@ -423,3 +423,14 @@ Two  <<<<----Correct
 - The defer attribute works only with promises.
 - The defer attribute will asynchronously load the scripts in order.   <<<<----Correct
 
+
+### Q50. The following program has a problem. What is it?
+```
+var a;
+var b = (a = 3) ? true: false
+```
+- The condition in the ternary is using the assignment operator.   <<<<----Correct
+- You can't define a variable without initializing it.
+- You can't use a ternary in the right-hand side of an assignment operator.
+- The code is using the deprecated var keyword
+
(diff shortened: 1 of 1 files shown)


flyingSquirrel-dev:
Just wondering the reason why commit bb569cd was not passed?

Ebazhanov:
> Just wondering the reason why commit [bb569cd](https://github.com/Ebazhanov/in-quiz-questions/commit/bb569cdaa008eea28762e8813e89135f5657b05f) was not passed?

yes, I would be happy if you can help with it 
https://github.com/Ebazhanov/in-quiz-questions/pull/305/checks?check_run_id=1001100849
I still was not able to find the solution

Ebazhanov (approved):
Great job! thanks for contributing!!!

Ebazhanov:
@all-contributors please add @flyingSquirrel-dev for design

Ebazhanov:
> I am not sure that the quiz no.50 has any problems to run.
> Code looks good, and JavaScript Engine is able to understand it I guess.
> 
> However, I think The condition in the ternary is using the assignment operator. is close to the correct answer because the others can not be the answer.
> 
> Please give me any feedbacks! :)

FYI have created post on SO regarding this question, hope we can get some feedback from there 
https://stackoverflow.com/questions/63485372/quiz-question-the-following-program-has-a-problem-what-is-it

flyingSquirrel-dev:
Wow, thanks :) I think the answer I chose is right :)

PR #347, opened 2020-09-08
Update java-quiz.md

lewing00:
I'm suggesting adding a "reasoning" or explanation for each question to help younger developers not only do well on these tests but understand these questions. I've added possible reasoning for the first question but would be happy to continue working on explanations for each question if approved.

### DIFF
--- a/java/java-quiz.md
+++ b/java/java-quiz.md
@@ -7,6 +7,8 @@ Java Assessment
 - awb
 - traw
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Good idea @lewing00! thank you for these improvements, and yes I think it could be really useful :heart_eyes:

Ebazhanov:
@all-contributors please add @lewing00 for design, ideas

PR #390, opened 2020-09-20
Fix a typo

vnborx:
The brackets are in the wrong place.


monkrus (commented):
Thanks for fixing a typo!

PR #430, opened 2020-09-30
fix: correct a wrong answer for Q30

DinaTaklit:
You've mistaken here once you margin vertically and u use margin top and margin bottom it will take the largest one not the sum up :p. Check here https://codepen.io/DinaTAKLIT/pen/eYZaqVw?editors=1100


monkrus (approved):

monkrus:
Thank you! @all-contributors please add @DinaTaklit for code

PR #431, opened 2020-09-30
fix: Correct a wrong answer for Q35

DinaTaklit:
Again, once we wante to evenly distribute the flex items we use space-arround value not space between. Check here: https://codepen.io/DinaTAKLIT/pen/KKzLOGJ

Please be careful you can always check the answers before deliver them. Some people take them as reference here :).

Thank you.
Best regards, 
Dina.


monkrus:
Thank you @DinaTaklit for the valuable input !

monkrus:
@all-contributors please add @DinaTaklit  for code

PR #434, opened 2020-10-01
Update c-sharp-quiz Docs. 

Dushyantsingh-ds:
Q38. What prints to the console when this code is executed?
public delegate void AuthCallback(bool validUser); 

public static AuthCallback loginCallback : Login; 

public static void Login() { 
Console.WriteLine("Valid user!"); 
} 
public static void Main(string[] args) {
 loginCallback(true); 
}

Correct Answer :  an error, because the method signature of Login doesn't match the delegate

Q39. How would you declare a sealed class named User?
Correct Answer : sealed class User

### DIFF
--- a/c#/c-sharp-quiz.md
+++ b/c#/c-sharp-quiz.md
@@ -330,13 +330,13 @@ loginCallback(true);
 
 - [ ] Login successful... 
 - [ ] Valid user! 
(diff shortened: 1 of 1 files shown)


monkrus (approved):
Nice!

monkrus:
@all-contributors please add @Dushyantsingh-ds for code

PR #472, opened 2020-10-03
Updated Correct answer for question 23

xybervenom:
Example Syntax-
class Foo():
        def __init__(self,x,y):
            print x+y
f = Foo(3,4)


monkrus (approved):

monkrus:
@all-contributors please add @xybervenom for code

PR #476, opened 2020-10-04
Marked answers of #Q44 to #Q48

goodday451999:


Ebazhanov (approved):
Very nice! You are so fast :runner: -> Approved!

Issue #531, opened 2020-10-10
OOP duplicate detected and some new q


ZdravekSprite:
Q40 = Q46

there is variation of Q23 [Two classes combine private data members and provide public member functions to access and manipulate those data member. Where is abstraction used?](https://youtu.be/dvgIYtR9wpc?t=82)

What type of inheritance may lead to the diamond problem?
[ ] single level
[ ] multilevel
[ ] hierarchical
[+] multiple
[Explanation: The diamond problem arises when multiple inheritance is used. This problem arises because the same name member functions get derived into a single class. Which in turn creates ambiguity in calling those methods.](https://www.sanfoundry.com/object-oriented-programming-questions-answers-multiple-inheritance/)

Which statement about compositions and aggregations is true?
[ ] If one element of aggregation is dereferenced, all its elements are eligible for garbage collection.
[ ] If a composition dies, the contents die.
[ ] An aggregation contains only abstract classes.
[ ] The contents of a composition are all siblings.
offered answers are the same as for Q37 but question is diffrent
BTW
Q37. How are contents of a composition different from those of aggregation?
- **if a composition dies, the contents die** <<-- correct [Programming Foundations: Object-Oriented Design|Inheritance and Composition|Q9/13](https://www.linkedin.com/learning/programming-foundations-object-oriented-design-3/quiz/5e0e855d498ee8fe91c8a01d?autoAdvance=false)

I had diffrend variant of one offered answer for Q3
Why would you create an abstract class, if it can have no real instances?
[ ] to explore a hypothetical class
[ ] to have common behavior in derived classes
[ ] to reserve memory for an unspecified class type
[ ] to prevent unwanted method implementation

and Q11
When is a constructor executed?
[ ] every time an object is referenced
[ ] when a class is defined using the `class` keyword
[ ] when an object is created from a class using the `create` keyword
[ ] when an object is created from a class

goodday451999:
@Ebazhanov I'm willing to handle this issue. Could you please allow me?

goodday451999:
@Ebazhanov please find the following updates-
- Replaced the duplicate question Q#46 with the new question "What type of inheritance may lead to the diamond problem?"
- The variation of the Q23 is very minor in the question ("if" is mentioned in the existing question where the provided question is not with that word. And there are no variations in the options), so I've not made any modifications regarding this issue.   
[...]

PR #536, opened 2020-10-12
Resolved the issue #531

goodday451999:
- Replaced the duplicate question Q#46 with the new question "What type of inheritance may lead to the diamond problem?"
- In case of having different variations of the existing questions (Q3, Q11, and Q37): the existing questions are numbered as Q3.1, Q11.1, and Q37.1, and the new variations are numbered as Q3.2, Q11.2, and Q37.2.


Ebazhanov (approved):
Really good idea! I like it ! thank you @goodday451999 :muscle:
Approved -> Merged

PR #576, opened 2020-10-15
Django

tik9:
After a while I am back with a commit;)


Ebazhanov (approved):
Very good! Thank you @tik9 Timo! You are making very clear and good PR as usual! :+1:

PR #632, opened 2020-10-23
Add CTE syntax for mysql

neslinesli93:
Add question about syntax for CTE (Common Table Expression)


Ebazhanov (approved):
Thanks Tommaso @neslinesli93, well done! :punch:

PR #656, opened 2020-10-27
Got answers to Questions 58-61

asarma2012:
Q58: Shows data being classified into more than two categories or classes. Thus, this is a multi-class classification challenge.
Q59: Underfitted data models usually have high bias and low variance. Overfitted data models have low bias and high variance.
Q60: The answer is self-explanatory: if Asian users are the only group of people making the complaint, then the training data should have more Asian faces.
Q61: This question is very similar to Q49 but involves a polar opposite scenario.


Ebazhanov (approved):
Thanks a lot for the PR @asarma2012, I also moved your explanations to the MD file for each answer, maybe it will be useful for other people! 

Ok, great will Approve and merge it!

Ebazhanov:
@all-contributors please add @asarma2012 for code, example

PR #668, opened 2020-10-29
Update ruby-on-rails-quiz.md

TaylorOD:
- formatting for question 13


Ebazhanov (approved):
No problem good catch! Approved

PR #676, opened 2020-10-30
Update adobe-illustrator-quiz.md

silverstroom:


Ebazhanov (approved):
That's great, I like this :punch:! Thank you @silverstroom

PR #680, opened 2020-10-30
Update README.md

TaylorOD:
- Change JavaScript to partially answered (only 1 question is answered currently)


Ebazhanov (approved):
You are the only person who is updating readme! Thank you @TaylorOD :+1:

PR #712, opened 2020-11-05
Update javascript-quiz.md

rradfar:
#### Fix / Add responses:
- 22 (corrected response & added explanation)
- 36 (corrected response & added explanation)
- 54 (removed confusing explanation)
- 60 (duplicate of Q.12 but worded differently so I left it alone in case a user googles the exact question and only added the correct response)

#### Remove duplicates:
- 15 & 23. Removed 23
- 18 & 32. Removed 32
- 45 & 58. Removed 58


Ebazhanov (approved):
@rradfar thank you for the detailed comments, and also very good improvements :punch:! Approved.

PR #754, opened 2020-11-13
T sql readme prettier

tik9:
As always, pleasure working with you Evgenii


Ebazhanov (approved):
For me as well :fist:, thank you @tik9, you are doing very good! :muscle:

PR #767, opened 2020-11-15
Fix error s swift-quiz. Add correct answer

amanbolat:
Q39: correct the answer [proof](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html#:~:text=A%20convenience%20initializer%20must%20call,ultimately%20call%20a%20designated%20initializer.)

Q29: range operator is `...`, not `..`


monkrus:
@all-contributors please add @amanbolat for design

PR #821, opened 2020-12-01
Add quiz 47 and update json quiz

DevHabit:
Updated json quiz and added quiz 47.


Ebazhanov (approved):
That's good that you are doing formatting in parallel together with the new question in one PR!
Have a great productive day @DevHabit!

PR #846, opened 2020-12-05
Answer Q48

AdrianBuenfil:
Adds answer for Q48. Tested on terminal.


Ebazhanov (approved):
Thank you for your explanation and that you are checking before creating PR :+1:

PR #851, opened 2020-12-07
Changed the option with the correct option

amitravikumar:
Q3. Which shell query displays all citizens with an age greater than or equal to 21?
db.citizens.select('WHERE age >= 21')
db.citizens.select('age >= 21')
db.citizens.select('WHERE age >= 21')
db.citizens.select ({age: {$qte: 21}}) <<<<<----CORRECT

The correct option for the following question is:
db.citizens.select ({age: {$gte: 21}}) <<<<<----CORRECT

Point updated is highlighted or bold marked.


Ebazhanov (approved):
Interesting, good catch! Thank you @amitravikumar :star:

Ebazhanov:
@all-contributors please add @amitravikumar for design

PR #853, opened 2020-12-07
Update python-quiz.md

schoppfe:
Q31. Given the following three list, how would you create a new list that matches the desired output printed below?
Trying the given solutions in Python 3.9.0 resulted in answer "B" as solution. Marked solution was corrected. 
The answer currently marked as correct results in an unprintable object: <zip object at 0x[...]>

### DIFF
--- a/python/python-quiz.md
+++ b/python/python-quiz.md
@@ -355,7 +355,7 @@ output.append(fruit_tuple)
 return output
 ```
 
-- [ ] `-`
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Ok, thank you @schoppfe! :+1:

Ebazhanov:
@all-contributors please add @schoppfe for design

PR #859, opened 2020-12-12
Removed null words. This happen only on Copy Paste

aamllc:


Ebazhanov (approved):
Agree with you :+1:! Thank you @aamllc, good catch :muscle:!

Ebazhanov:
@all-contributors please add @aamllc for design

PR #869, opened 2020-12-16
Q33. correct answer - start

pravin-sharma:
refer docs https://nodejs.org/api/events.html


monkrus:
@all-contributors please add @pravin-sharma for design

PR #875, opened 2020-12-16
Update html-quiz.md

spitlisk:


Ebazhanov (changes requested):
I think and quite sure that before it was more informative and visible, but with your link it become more compact. I am thinking about to keep both your link and previous solution (table). What do you think @spitlisk? and also please use special markdown format for posting the link like `[](url)`

spitlisk:
> I think and quite sure that before it was more informative and visible, but with your link it become more compact. I am thinking about to keep both your link and previous solution (table). What do you think @spitlisk? and also please use special markdown format for posting the link like `[](url)`

I put the link of the image to someone for the document like question 12 I don't know how to do it, I can learn.

Ebazhanov:
> > I think and quite sure that before it was more informative and visible, but with your link it become more compact. I am thinking about to keep both your link and previous solution (table). What do you think @spitlisk? and also please use special markdown format for posting the link like `[](url)`
> 
> I put the link of the image to someone for the document like question 12 I don't know how to do it, I can learn.

Thanks would be great and ping me later when should I take a look again! Have a good day!

Ebazhanov (changes requested):
Hello, @spitlisk! Please let me know if you need any help with this PR, waiting for your updates.

PR #878, opened 2020-12-16
Update html-quiz.md

spitlisk:


html/html-quiz.md
```diff
@@ -556,8 +556,10 @@ From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr): The HT
 --Paper
 -Groceries
 --Milk
+https://cdn.lynda.com/question/5d07ce7b498ef292ca309a14/5d07ce7b498ef292ca309a14-1560793406126
```
Ebazhanov, line 559: you can add like this (it will be displayed as a hyperlink) and give the name FYI @spitlisk 
```suggestion
(some_name)[https://cdn.lynda.com/question/5d07ce7b498ef292ca309a14/5d07ce7b498ef292ca309a14-1560793406126]
```

Ebazhanov (changes requested):

html/html-quiz.md
spitlisk: thanks for the tip

PR #951, opened 2021-01-07
Readme and c++

tik9:


tik9:
Again that much commits here. I have to reduce it to one commit which is "commit from tik9".

tik9:
Can merge;)

PR #961, opened 2021-01-09
Updated a typo

rzamora11:
Which "story" about technical debt is true? Is incorrect. It is Which "statement" about technical debt is true.


Ebazhanov (approved):
Super, thank you @rzamora11! Merged

PR #978, opened 2021-01-13
Update reactjs-quiz.md Q:48

BegovicNikola:
- Every property you add to component in view will be added to props object (since the functional component doesn't hold state props are only source of truth);
- Functional component is basically a JavaScript/ES6 function that returns a React element (JSX);

### DIFF
--- a/react/reactjs-quiz.md
+++ b/react/reactjs-quiz.md
@@ -466,10 +466,10 @@ Currently, `handleClick` is being called instead of passed as a reference. How d
 - [ ] `<button onClick={this.handleClick}>Click this</button>` 
 - [ ] `<button onclick={this.handleClick}>Click this</button>`
 
-#### Q48. (Not sure answer)
+#### Q48.
 Which answer best describes a function component?
 - [ ] `A function component is the same as a class component.`
-- [ ] `A function component accepts a single props object and returns a React element.`
+- [x] `A function component accepts a single props object and returns a React element.`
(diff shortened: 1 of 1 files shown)


react/reactjs-quiz.md
```diff
@@ -466,10 +466,10 @@ Currently, `handleClick` is being called instead of passed as a reference. How d
 - [ ] `<button onClick={this.handleClick}>Click this</button>` 
 - [ ] `<button onclick={this.handleClick}>Click this</button>`
 
-#### Q48. (Not sure answer)
+#### Q48.
 Which answer best describes a function component?
 - [ ] `A function component is the same as a class component.`
-- [ ] `A function component accepts a single props object and returns a React element.`
+- [x] `A function component accepts a single props object and returns a React element.`
```
Ebazhanov, line 472: @BegovicNikola thanks for the answer, but could you please add an explanation right after

 something like this:
```suggestion
- [x] `A function component accepts a single props object and returns a React element.`
// Every property you add to component in view will be added to props object (since the functional component doesn't hold state props are only source of truth);
Functional component is basically a JavaScript/ES6 function that returns a React element (JSX);
```
Just to make it more useful for others, thank you in advance!

Ebazhanov (changes requested):
FYI @BegovicNikola left some comments

Ebazhanov:
@all-contributors please add @BegovicNikola for code, design

PR #979, opened 2021-01-13
Nosql readme

tik9:
This is pulled without rebase parameter. Look at commit c4294fb (last one in commit history).


Ebazhanov:
> This is pulled without rebase parameter. Look at commit [c4294fb](https://github.com/Ebazhanov/linkedin-quiz-assessments/commit/c4294fbc2b982c570e58a51ea16dc5954782d132) (last one in commit history).

good one 👍 , but still we should think why there are some other changes
image

PR #980, opened 2021-01-13
Azure

tik9:
This is pulled with rebase parameter. Look at commit branch azure , last one in commit history. Did not work either, still too much commits.


Ebazhanov:
> This is pulled with rebase parameter. Look at commit branch azure , last one in commit history. Did not work either, still too much commits.

yes, agree

PR #1021, opened 2021-01-22
Fixed wrong answers in T-SQL: Q10, Q27, Q28

Basim108:
Q10 - the picture of table is important here and there it can be seen that there are only two value with min points. Secondly, the previous answer was wrong because order by DESC will put highest points into the beginning of result list and TOP(2) will take first two highest points, and we need the lowest points.

Q27 - top(1) with ties will take the highest grade and all other students with the same grade (because they are order by grade) and matches the highest grade.

Q28 - select * from dbo.books  here dbo is a schema and the inventory is also schema; if we'd like to specify a database we should use db_name.schema_name.table_name


Ebazhanov (approved):
Thank you @Basim108 very valuable contribution :muscle:! And I promised to add your explanation inside the *.md file later!

Ebazhanov:
@all-contributors please add @Basim108 for design, code

PR #1124, opened 2021-02-11
Create LinkedIn mongo quiz.md

hp-techbucket:
some quiz questions from LinkedIn


Ebazhanov (changes requested):
@hp-techbucket looks good :+1: but can you please move your new questions into already existed [quiz](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/master/mongodb/mongodb-quiz.md), maybe we can keep them there somehow! But in general all good! Let me know if you have any other questions, I am always here! Have a good day!

PR #1131, opened 2021-02-11
Update html-quiz.md

AirlineDog:
Fixed some questions to match linkedin quiz's.
Added two more questions with answers
Deleted a duplicate question

### DIFF
--- a/html/html-quiz.md
+++ b/html/html-quiz.md
@@ -28,7 +28,7 @@
 - [ ] as children of `<body>, <article>, <aside>, <nav>, and <section>` tags
 - [ ] as children of `<body>, <article>, <table>, and <section>` tags
 
-#### Q5. What's the best way to apply bold styling to text?
+#### Q5. What is the best way to apply bold styling to text?
 
 - [x] `<strong>`
 - [ ] Use CSS.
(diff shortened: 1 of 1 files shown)


Ebazhanov (changes requested):
@AirlineDog very nice but I want you revert back question 
`#### Q36. What code will produce this table?` because I found that it was not duplication.
Thank you in advance!

AirlineDog:
@Ebazhanov Αre you sure that Q36 is not an incomplete duplicate of Q26, because i can find no difference except that Q36 has 3 answers and Q26 has four.
If you that is not the case here, I will be happy to revert the question back.

Ebazhanov:
> @Ebazhanov Αre you sure that Q36 is not an incomplete duplicate of Q26, because i can find no difference except that Q36 has 3 answers and Q26 has four.
> If you that is not the case here, I will be happy to revert the question back.

Ahh, yes! You are right the questions inside the tests Q26 are similar to Q36 true! Maybe Q36 has new version so, nice catch once again! Could you please fix conflicts, and we can merge your branch! Thank you in advance @AirlineDog!

image

AirlineDog:
@Ebazhanov Conflicts Resolved.

Ebazhanov (approved):
You are amazingly fast :runner:! Thank you for this contribution! :punch:

PR #1145, opened 2021-02-13
OOP CRC

Behnam-Emamian:


Ebazhanov (approved):
Also, great that you answered it here! Thank you @Behnam-Emamian!

PR #1241, opened 2021-03-05
Update objective-c-quiz.md

nullfx:
- Answered a few questions
- Corrected one or two
- Removed one that was repeated later on. Question 21 had the wrong question text and was repeated correctly as 32.
- Added 3 new questions that weren't here when i took it

One question remains unanswered ( now `Q32 What is the type of the error return value stored in json` which may be transcribed incorrectly `json` is an `id` but the question may be asked incorrectly)


Ebazhanov (approved):
Well done @nullfx, and sorry for delay with approve! :punch:

Ebazhanov:
@all-contributors please add @nullfx for design, code

PR #1244, opened 2021-03-06
Update java-quiz.md

KidusMT:
update on the java question #7 answer which was a wrong answer.
```
jshell> str.substring(3,4);
$1 ==> "d"
```

### DIFF
--- a/java/java-quiz.md
+++ b/java/java-quiz.md
@@ -83,8 +83,8 @@ System.out.println(str);
 
 - [ ] CD
 - [ ] CDE
-- [ ] D
-- [x] "abcde"
+- [x] D
+- [ ] "abcde"
 
 #### Q8. What is the result of this code?
 


rradfar:
If we try try running this piece of code, it outputs 'abcde'. I believe the reason for this behavior is that the 'toUpperCase()' and 'substring()' methods do not mutate the original string which is what we are printing at the end.

```java
String str = "abcde";
str.trim();
str.toUpperCase();
str.substring(3, 4);
System.out.println(str);  //abcde
```

If I recall correctly, strings in Java are not mutable so any operation on a string, just returns a new string.

KidusMT:
Correctly put @rradfar . I only tested it on jshell for that segment of code. I didn't try to "System.out.println" it.. so yes.correct. Discard this pull request.

PR #1272, opened 2021-03-11
Update html-quiz.md

itware:
Both the **b** and **strong** tags have the same functioning, but **strong** tag additionally adds semantic strong importance o the text. 
But we have to apply bold styling only.

**added**
**Ooops.... just passed the quiz, and **b** option now is absent. So, **strong** is the best choice.**


rradfar:
Should we close this PR then? 😄

itware:
sure)

PR #1275, opened 2021-03-11
Update vba-quiz.md

IamNotPeterPan:


lnugraha (approved):
Require minor formatting for question #12, such that the reference does not clump with the last choice of question 12

Ebazhanov:
> Require minor formatting for question #12, such that the reference does not clump with the last choice of question 12

Agree with your comment @lnugraha, you are fully right 💯 , but it's fine I can do it by myself since `IamNotPeterPan` is unavailable.

PR #1357, opened 2021-03-30
Update python-quiz.md

nicktids:
I did a search and  for 5! = 6 as there were spaces, and didn't find it because we had here no gaps 

I know it's like the easiest quesiton but might stumble some people.


Ebazhanov (approved):
You are right @nicktids your changes is very important even they look minor! Good job, thank you! 👍

Ebazhanov:
@all-contributors please add @nicktids for design

PR #1392, opened 2021-04-06
Corrected wrong answer on Q44

D4RIO:
Actually, there is an error on every line because of the "bash" at the beggining, but that seems just a mistake.
The proper bash syntax would be: for i in $(ls); do ... ; done


Ebazhanov (approved):
It seems to me that you are an expert in `bash`, thank you for your PR @D4RIO! 👍

PR #1393, opened 2021-04-06
Fixed wrong answer to Q15

D4RIO:
tee - read from standard input and write to standard output and files


Ebazhanov (approved):
Thank you, @D4RIO, your correction are always welcome! Have a good day!

PR #1394, opened 2021-04-06
Fixed wrong answer to Q28

D4RIO:
This behaviour is controlled by the SUID permission. You can see the SUID mode for passwd with the command below:

   ls -l $(whereis -b passwd | sed s,^.*:,,)

Here, you will see 'rws' for the owner, which is root. About the 's', the info pages for ls are helpful:

     ‘s’
          If the set-user-ID or set-group-ID bit and the corresponding
          executable bit are both set.

     ‘S’
          If the set-user-ID or set-group-ID bit is set but the
          corresponding executable bit is not set.

### DIFF
--- a/linux/linux-assesment.md
+++ b/linux/linux-assesment.md
@@ -208,8 +208,8 @@ echo \$myNumber | sed -e 's/^[[:digit:]][[:digit:]][[:digit:]]/(&)/g'
 #### Q28. Why is the passwd command able to modify the /etc/passwd file?
 
 - [ ] The kernel identifies it as one of the names of extended commands.
-- [x] It has the same name as the file it modifies.
-- [ ] It has the SUID permission mode and is owned by root.
+- [ ] It has the same name as the file it modifies.
+- [x] It has the SUID permission mode and is owned by root.
 - [ ] It is a system administration command.
 
 #### Q29. When a user deletes a file using the rm command, Linux will **\_**.


Ebazhanov (approved):
@D4RIO thanks a lot, FYI I am not an expert of Linux (using Mac), so I like your description very much and maybe next time you can add it directly inside the question just to keep it for other people! Have a good day! :hibiscus:

Ebazhanov:
@D4RIO have the same photo from the past
image

D4RIO:
> @D4RIO thanks a lot, FYI I am not an expert of Linux (using Mac), so I like your description very much and maybe next time you can add it directly inside the question just to keep it for other people! Have a good day! 🌺

I have been considering to open a discussion about this, because it depends on the way people use this repo. Right now, maybe this is just to get the answers for the provided questions, but maybe with little descriptions or some links it can also provide some insight. Also, bringing clarity to questions that require some detailed comprehension on the subject may avoid misinterpretation and PRs bringing the same wrong answers. I think that, if the community wants it, it can also be a suggestion added to the Contributing file, so every new contributor will feel free to include some details, and this will help us find out the better answer as well.

Ebazhanov:
I'm not sure what kind of community we have, usually people come here to find answers to their questions and pass the test successfully! And you are right about the answers that can change from right to wrong people when the test has no description or reference link. Any suggestion or inputs are very welcome!

Ebazhanov:
Feel free to update contributing file (thank you in advance!)

D4RIO:
> @D4RIO have the same photo from the past
> image

Your camera is beautiful. I love it. Seems, mostly because of the type of noise, to be film (not a digital camera). I'm a big enthusiast about photography. As there is a deep interaction between physics and art, photography could be considered a kind of engineering, and it is as fascinating as any other, and a bit more too, just like music.

Ebazhanov:
Totally agree with your thoughts, I was born when there were no mobile phones and laptops yet, that's why I was fond of black and white photography

PR #1402, opened 2021-04-06
Fixed Q22 on bash-quiz.md

D4RIO:
The original answer interpreted that "find . -name *.html" is okay too, but the shell will try to expand the glob * before running find.
You can find out more about this with "man bash", under EXPANSION/Pathname Expansion.

### DIFF
--- a/bash/bash-quiz.md
+++ b/bash/bash-quiz.md
@@ -182,11 +182,12 @@ E. bash demo.sh
 
 #### Q22. How could you get a list of all .html files in your tree?
 
-#### Idk both of them work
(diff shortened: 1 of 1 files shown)


D4RIO:
> The original answer interpreted that "find . -name *.html" is okay too, but the shell will try to expand the glob * before running find.
> You can find out more about this with "man bash", under EXPANSION/Pathname Expansion.

I don't know if

Ebazhanov (approved):
Thanks, @D4RIO, cool! And great that you left some comment inside the `*.md` file 👍

PR #1404, opened 2021-04-07
Professional in Human Resources PHR linkedin

ANC4DES:
visit below link for all exam answer with complete 100%
https://www.youtube.com/watch?v=6F-Ms_piXKI
Video recorded by me.


Ebazhanov (changes requested):
@ANC4DES great that you recorded the video I like it, but there is nothing to merge I think we should create an issue and let other people do the rest. What do you think? If you need help please let me know.

Issue #1400, opened 2021-04-06
Machine Learning 


shay144:
I think qst 1 in machine learning should be K-means

PR #1407, opened 2021-04-07
Update answer for qn. 01

RiyaMathew-11:
I think the correct answer for qn. 1 is the K-means cluster algorithm. The reference was also given in issue #1400. So made the correction.

### DIFF
--- a/machine-learning/machine-learning-quiz.md
+++ b/machine-learning/machine-learning-quiz.md
@@ -2,10 +2,10 @@
 
 #### Q1. You are part of data science team that is working for a national fast-food chain. You create a simple report that shows trend: Customers who visit the store more often and buy smaller meals spend more than customers who visit less frequently and buy larger meals. What is the most likely diagram that your team created?
 
-- [x] multiclass classification diagram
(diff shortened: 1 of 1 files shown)


D4RIO (commented):
Please, would you check Q2? I think you introduced a small change, like a blank space or so.

RiyaMathew-11:
> Please, would you check Q2? I think you introduced a small change, like a blank space or so.

Sure, I would check it out. Thanks for pointing out.

RiyaMathew-11:
> > Please, would you check Q2? I think you introduced a small change, like a blank space or so.
> 
> Sure, I would check it out. Thanks for pointing it out.
It's a small change in whitespace. I've compared with both. It doesn't  show up in the content

RiyaMathew-11:
It's a small change in whitespace. I've compared with both. It doesn't show up in the content

Ebazhanov (approved):
Well done @RiyaMathew-11 :hibiscus:, thanks for the answer! 👍

Ebazhanov:
@all-contributors please add @RiyaMathew-11 for code, design

PR #1414, opened 2021-04-08
Fixed Q31 in Linux

D4RIO:
Answer said that brace expansion requires files to exist.
You can test this in a shell by typing `echo number{1,2}`. Parameters expand to `number1` and `number2` always. Brace expansion creates a list.
Now try with globs and type `echo unreal*file`. If the file doesn't exist, and nullglob is not set, then it will remain the same. If there is at least one file matching the pattern, it will expand to the list of matching pathnames.


Ebazhanov (approved):
Great input 👍  @D4RIO, waiting when you will add explanation into this question! Thank you in advance!

Issue #1309, opened 2021-03-22
Python - What is a lambda function?


phmonteiro:
[ ] any function that makes use of scientific or mathematical constants, often represented by Greek letters in academic writing
[ ] a function that get executed when decorators are used
[ ] any function whose definition is contained within five lines of code or fewer
[ ] a small, anonymous function that can take any number of arguments but has only expression to evaluate

https://imgur.com/a/27Hw2jP

lnugraha:
**Answer:**
[ ] any function that makes use of scientific or mathematical constants, often represented by Greek letters in academic writing
[ ] a function that get executed when decorators are used
[ ] any function whose definition is contained within five lines of code or fewer
[x] a small, anonymous function that can take any number of arguments but has only expression to evaluate

**Explanation:** the lambda notation is basically an anonymous function that can take any number of arguments with only single expression (i.e, cannot be overloaded).

It has been introducted in other programming languages, such as C++ and Java. The lambda notation allows programmers to "bypass" function declaration.

RiyaMathew-11:
I think this issue could be closed now

Ebazhanov:
Thank you, @lnugraha @phmonteiro!👍  I will update quiz regarding your questions today!

PR #1423, opened 2021-04-08
Update python - added question on lambda functions with Reference

RiyaMathew-11:
In response to issue #1309 - updated the question on lambda functions in python and added reference


Ebazhanov (approved):
Thank you :star: for your PR @RiyaMathew-11, also if you think issue may be closed, please do so. Thank you in advance! :tulip:

PR #1433, opened 2021-04-11
Update android-quiz.md

Parveshdhull:
19. In method "getDefaultSharedPrefarances(this).getString()" Second parameter is passed so that it can be returned, in case key doesn't exist. If we pass null as second parameter, then it will return null in case of null. Which we were trying to avoid

37. Intent filters are used to make activities accessible to other apps using intents. So we have to choose option which have no intent filter to make sure it is not accessible by intent

### DIFF
--- a/android/android-quiz.md
+++ b/android/android-quiz.md
@@ -197,6 +197,8 @@
 - [x] getDefaultSharedPreferances(this).getString(EMAIL,"")
 - [ ] getPreferances(this).getString(EMAIL,null)
 
+**Explanation:** In Method "getDefaultSharedPrefarances(this).getString()" Second parameter is passed so that it can be returned, in case key doesn't exist. So we need to pass an empty string to be returned in case key doesn't exist.
+
 #### Q20. Why is it problematic to define sizes using pixels on Android?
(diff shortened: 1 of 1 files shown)


Parveshdhull:
First question No is 19, But second is 37. I don't know why github is showing it as 20 while I typed 37

image

Ebazhanov:
> 1. In method "getDefaultSharedPrefarances(this).getString()" Second parameter is passed so that it can be returned, in case key doesn't exist. If we pass null as second parameter, then it will return null in case of null. Which we were trying to avoid
> 2. Intent filters are used to make activities accessible to other apps using intents. So we have to choose option which have no intent filter to make sure it is not accessible by intent

Hello, @Parveshdhull! Thanks a lot for your PR, I would suggest you to put your explanation inside *.md file because this will be useful in the future for other people! What do you think?

Parveshdhull:
Hey @Ebazhanov , I added explanation in android-quiz.md next to questions

Ebazhanov (approved):
Cool, thanks again @Parveshdhull for your valuable contribution! 👍

Ebazhanov:
@all-contributors please add @Parveshdhull for code, design

PR #1445, opened 2021-04-13
aws Q55

gudata:
The outbound rules for the http 80 must be allowed also.


Ebazhanov (approved):
Another good implements :muscle: @gudata thanks!

davebaol:
This doesn't make any sense to me.
Being the instance a web server, you need an inbound rule to allow for HTTP traffic to reach the instance.
This woud also make response traffic implicitly allowed because security groups are stateful.

The least worst of the four answers presented here is the third.
However it does not resolve the reachability of the server over http.

I honestly think there is something wrong either with the question or the answers.

Ebazhanov:
Hello, @davebaol! Thanks for your feedback 👍 , feel free to create pull request!

PR #1446, opened 2021-04-13
Added two more questions about wordpress and security

gudata:


Ebazhanov (approved):
You are working very hard :heart:, I wish you a very good day @gudata !

PR #1500, opened 2021-04-28
Fix : mark down and checkbox

sidisan:
fixing checkbox that not shown and formating some mardown


Ebazhanov (approved):
Good job 👍 ! Thank you @sidisan!

PR #1527, opened 2021-05-08
fix the maximum number of columns that can be used by a single table index

narutubaderddin:


Ebazhanov:
@all-contributors please add @narutubaderddin for code, design

PR #1528, opened 2021-05-08
Q83. background-position

narutubaderddin:
The first value is the horizontal position and the second value is the vertical. The top left corner is 0% 0%. The right bottom corner is 100% 100%.


Ebazhanov (approved):
Very nice :heart_eyes:, thanks @narutubaderddin! Have a great day!

PR #1536, opened 2021-05-11
Add more questions to REST API (Q49 - Q51)

meghatiku:
Added 4 more questions for REST API topic with answers. 

Kindly check the correctness of answers as well. TIA 👍🏻

PR #1540, opened 2021-05-12
Update question count for REST API in readme

meghatiku:
Needed to update readme after https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/1536


Ebazhanov (approved):
oh yes, good catch @meghatiku 👍  Approved and merged!

PR #1569, opened 2021-05-18
Updated JSON Quiz Q2 - Fixes #1568

### LINKED ISSUE #1568: JSON Q2
In JSON Q2, `Which code is the closest JSON equivalent of the data shown?` the correct option marked is

` {
    "color": "red",
    "color": "blue",
    "color": "green"
  }
`

but according to the question, the given data is an array of strings, and I believe the correct option should be 

`
[
    "red",
    "blue",
    "green"
  ]
`

as it is a valid JSON (ref attached)

image

iHamzaKhanzada:
Fixed incorrect answer for Q2


Ebazhanov (approved):
Good catch @iHamzaKhanzada, thanks for your input :muscle:

Ebazhanov:
@all-contributors please add @iHamzaKhanzada for code

PR #1586, opened 2021-05-21
Add three problems

pro335:


Ebazhanov (approved):
New questions and answers are always very welcome :punch:! Thanks a lot @pro335 🥇

PR #1630, opened 2021-05-31
Removed duplicates on C++ sheet

danie007:
- Removed duplicate questions & fixed numbering
- Formatted code with added reference


Ebazhanov (approved):
Well come back @danie007, thank you 👍 !

PR #1644, opened 2021-06-02
Answered Hadoop Q30

VishwaSangani:
map function processes a certain key-value pair and emits a certain number of key-value pairs and the Reduce function processes values grouped by the same key and emits another set of key-value pairs as output.


Ebazhanov (approved):
Thanks a lot @VishwaSangani 🥇 , I just add your explanation under the answer!

Ebazhanov:
@all-contributors please add @VishwaSangani for design, code

PR #1690, opened 2021-06-10
fix Cpp q64

vmicho:
fix copy-pase ^.^


Ebazhanov (approved):
Thanks @vmicho, good catch 💯 !

PR #1706, opened 2021-06-15
Updated Answer for Q.39

SaqibAlMalik:
Absolute estimating is more reliable than relative estimating is not correct


Ebazhanov (approved):
Thanks for correcting @SaqibAlMalik 👍

PR #1737, opened 2021-06-22
Update questions

islamarr:
- Q49 is duplicated with Q41
- Q51 is duplicated with Q9
- I answered Q50


Ebazhanov (approved):
Thanks again @islamarr, you are doing great job! :pray:

PR #1747, opened 2021-06-25
Solidworks number of questions

djayor:
Instead of 29 questions and answers, Solidworks just have 3.


Ebazhanov (approved):
Good catch 👍 , thank you @djayor 💯

Ebazhanov:
@all-contributors please add @djayor for design

PR #1769, opened 2021-06-28
one fix and added q58~62

bloodiator2:


Ebazhanov (approved):
An other great PR, thank you @bloodiator2 :punch:

PR #1771, opened 2021-06-28
added Q58~71 (come on guys, why so many left-out)

bloodiator2:


Ebazhanov (approved):
Wow so many new questions and answers you are the greatest contributor, thanks a lot!

PR #1789, opened 2021-07-01
Corrected the answers of Q64, Q66, Q71 and  Q72

gazihasanrahman:
Q63: The answer should be "=VLOOKUP(J2,A$2:C$7,3,FALSE)", because we are interested in the value of the 3rd column of the table
Q64: The answer should be "=SUM(C15:Y15)", the sum of columns C to Y for the same row 15
Q66: The answer should be "Select `Paste Special > Transpose", because it needs to be transposed without creating a reference
Q71: The answer should be "=SUM($B2:B2)", we are calculating the running total here. Tested and worked
Q72: The answer should be "3", 27/6 = 4, and the reminder is 3

### DIFF
--- a/microsoft-excel/microsoft-excel-quiz.md
+++ b/microsoft-excel/microsoft-excel-quiz.md
@@ -477,15 +477,15 @@
 ![image](https://i.imgur.com/SAJTiQU.png)
 
 - [ ] `=VLOOKUP(A2,J2:L5,1,FALSE)`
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Amazing staff :heart_eyes:, thanks for your pull request @gazihasanrahman 👍 , going to add your explanation inside md file.

Ebazhanov:
@all-contributors please add @gazihasanrahman for design, code

PR #1831, opened 2021-07-11
Update xml-quiz.md

anandnat:
Added answers for Q20, Q21, Q22, Q32 & Q36

### DIFF
--- a/xml/xml-quiz.md
+++ b/xml/xml-quiz.md
@@ -206,7 +206,7 @@
 
 #### Q20. You want to convert a large XML file into CSV format. You did not create the XML file, so you are not familiar with all of the syntax. What will help you get the best insight into the file contents?
 
-- [ ] XSLT
+- [x] XSLT
 - [ ] DOM
 - [ ] AJAX
 - [ ] XSD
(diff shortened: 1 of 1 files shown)


xml/xml-quiz.md
```diff
@@ -206,7 +206,7 @@
 
 #### Q20. You want to convert a large XML file into CSV format. You did not create the XML file, so you are not familiar with all of the syntax. What will help you get the best insight into the file contents?
 
-- [ ] XSLT
+- [x] XSLT
```
ruslanbes, line 209: XSLT is a transformation process. But the question is not about the transformation, it's about understanding the XML. XSD is the correct one here - that's the schema document, that describes the XML.

ruslanbes (changes requested):

xml/xml-quiz.md
Ebazhanov: Thanks for review @ruslanbes I'll update the answer and put your explanation in *md file

Ebazhanov (approved):
Thank you @anandnat, for add new answers, great job! :punch:

PR #1858, opened 2021-07-16
Update html-quiz.md

Muhammad-Osama-Qureshi:
79 and 84 correct answers

### DIFF
--- a/html/html-quiz.md
+++ b/html/html-quiz.md
@@ -1264,10 +1264,9 @@ As Steve Krug once said, happy talk must die.
     6640 via Real <br>
     <a href="mailto:info@linkedin.com">info@linkedin.com</a>
   </em></p>
-</p>
 ```
 
-- [ ]
+- [X]
 
 ```markdown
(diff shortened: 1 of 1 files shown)


html/html-quiz.md
```diff
@@ -1388,5 +1387,5 @@ As Steve Krug once said, happy talk must die.
       ![B](images/Q84-2.jpg)
 - [ ] C
       ![C](images/Q84-3.jpg)
-- [ ] D
+- [X] D
```
ruslanbes, line 1390: B is correct here. (Show only "> Details"). Tested in both Firefox and Chrome

ruslanbes (changes requested):

html/html-quiz.md
Ebazhanov: @Muhammad-Osama-Qureshi thanks for your input and could you please take a look at @ruslanbes suggestion
would be happy to merge your PR
have a great day!

PR #2010, opened 2021-07-31
Answered Q45

AbhiBiju:
Source: [W3Schools](https://www.w3schools.com/css/css_positioning.asp)

```
position: absolute;
An element with position: absolute; is positioned relative to the nearest positioned ancestor (instead of positioned relative to the viewport, like fixed).

However; if an absolute positioned element has no positioned ancestors, it uses the document body, and moves along with page scrolling.
```
Screen Shot 2021-07-31 at 12 58 04 PM

### DIFF
--- a/front-end-development/front-end-development-quiz.md
+++ b/front-end-development/front-end-development-quiz.md
@@ -429,16 +429,20 @@ console.log(currencies);
 - [ ] `<input type="radio" value="vanilla" on>`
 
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thanks a lot @AbhiBiju, lovely pull request :heart:! Approved and merged :point_left:

Ebazhanov:
@all-contributors please add @AbhiBiju for design, code

PR #2013, opened 2021-08-01
Removed 4 duplicated questions in Python

Romazho:
In Python,
I removed Q50 which is the same as Q34 and 
I removed Q39 which is the same as Q33 and
I removed Q59 which is the same as Q52 and
I removed Q79 which is the same as Q36.
I am referring the question numbers in master, not in my PR :).


Ebazhanov (approved):
You have very sharp eyes, thank you @Romazho 👍  for your great improvements and valuable contribution! :muscle:

PR #2020, opened 2021-08-03
Update python-quiz.md

joaocasarin:
Added Q85.
When instantiating a new object from a given class, the `__init__()` method will take both `attr1` and `attr2`, and set its values to their corresponding object attribute, that's why the need of using `self.attr1 = attr1` instead of `attr1 = attr1`.


Ebazhanov (approved):
Hello, @joaocasarin! Thank you for your valuable input and explanation! :punch: 🥇

Ebazhanov:
@all-contributors please add @joaocasarin for design, code

PR #2040, opened 2021-08-08
Update kotlin-quiz.md

imDevSalman:
added two new questions, if someone can cross check the answers to those that would be great.


Ebazhanov (approved):
Thanks for new questions with answers @imDevSalman it is definitely great improvement, regarding to the answers, I have no idea who can check. But I believe your answers is 💯 correct!!!!! 🥇

Ebazhanov:
@all-contributors please add @imDevSalman for design, code

PR #2043, opened 2021-08-09
Update greetings.yml

CrankyEnemy:
Update PR message


Ebazhanov:
Thanks @CrankyEnemy 👍  but let's postpone this update, does not make sense to have it right now!

PR #2097, opened 2021-08-23
Delete duplicate question Q8 vs Q18

ytliuSVN:


Ebazhanov (approved):
Well done @ytliuSVN :punch:

PR #2136, opened 2021-08-29
Answer Q48 and add W3Schools source/explanation

AbhiBiju:
When you have a float property on an element and you want the next element to be below it, the clear property has to be used, matching the clear property to the float property. So if the image had a float: right; then the next element should have a matching clear: right;
Screen Shot 2021-08-28 at 7 59 29 PM


Ebazhanov (approved):
Thanks @AbhiBiju for adding answer with reference link and explanation, great job! 👍 🥇

PR #2141, opened 2021-08-30
Add new Linux question with answer

ytliuSVN:
Single dashes `-` for single-character options
double dashes `--` for words


Ebazhanov (approved):
LGTM 💯  thank you @ytliuSVN

PR #2165, opened 2021-09-03
Q20, Q32 Correct Answers

bodi1981:
Q20 https://docs.microsoft.com/en-us/sql/t-sql/statements/create-schema-transact-sql?view=sql-server-ver15
Q32 Column "Students.Grade" is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause.


Ebazhanov (approved):
Thanks @bodi1981 for the reference link and explanation :muscle:

PR #2169, opened 2021-09-04
Update python-quiz.md

DevaraatJoshi23:
The 10th question given here is now asked as 

Q10. What is the correct syntax for defining a class called Game, if it inherits from a parent class called LogicGame?

- [x] `class Game(LogicGame): pass`
- [ ] `def Game(LogicGame): pass`
- [ ] `def Game.LogicGame(): pass`
- [ ] `class Game.LogicGame(): pass`

Explanation : The parent class which is inherited is passed as an argument to the child class. Therefore, here the first option is the right answer.

### DIFF
--- a/python/python-quiz.md
+++ b/python/python-quiz.md
@@ -72,6 +72,8 @@
 - [ ] `def Game.LogicGame(): pass`
 - [ ] `class Game.LogicGame(): pass`
 
+**Explanation:** `The parent class which is inherited is passed as an argument to the child class. Therefore, here the first option is the right answer.`
+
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thank you @DevaraatJoshi23 for your correction and adding explanation! It is very valuable :fire:

Ebazhanov:
@all-contributors please add @DevaraatJoshi23 for design, code

DevaraatJoshi23:
Thank you @Ebazhanov

PR #2172, opened 2021-09-05
Update #### Q79 with right answer

rogue-coder:
Updating #### Q79. Which operator is used to concatenate Strings in Java with right answer and reference


monkrus (approved):

monkrus:
Thank you for updating! @all-contributors please add @rogue-coder  for code

PR #2239, opened 2021-09-17
Update rest-api-quiz.md

karuissobusy:
#### Q25. Which OAuth grant type can support a refresh token?
The answer is Authorization code grant that supports refresh tokens optionally. Reference: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.3


Ebazhanov (approved):
Thanks a lot @karuissobusy, good job :muscle:! I also added your reference link under the question!

Ebazhanov:
@all-contributors please add @karuissobusy code

PR #2250, opened 2021-09-19
Update Excel Questions

parker-stephens:
Only question I am unsure about is 92, but I'm leaning toward the last answer.


Ebazhanov (approved):
Great job anyway, thank you @parker-stephens! 👍

PR #2342, opened 2021-10-04
Update php-quiz.md

mazedlx:
Removed the note regarding the output of Q57. It returns 13, no matter the PHP Version (Testet with 4.4.9 - 8.0.10)


Ebazhanov (approved):
Good catch 👍 , thank you @mazedlx!

PR #2345, opened 2021-10-05
Update swift-quiz.md

baronfac:
Q17 did have a wrong answer. It was ["BEAR", "JOE", "CLARK"], which is wrong. names is a constant in this case thus it does not change

#### Q17. What is the value of names after this code is executed?

```swift
let names = ["Bear", "Joe", "Clark"]
names.map { (s) -> String in
  return s.uppercased()
}
```

- [ ] ["BEAR", "JOE", "CLARK"]
- [ ] ["B", "J", "C"]
- [x] ["Bear", "Joe", "Clark"]   
- [ ] This code contains an error.

### DIFF
--- a/swift/swift-quiz.md
+++ b/swift/swift-quiz.md
@@ -189,9 +189,9 @@ names.map { (s) -> String in
 }
 ```
 
-- [x] ["BEAR", "JOE", "CLARK"]
+- [ ] ["BEAR", "JOE", "CLARK"]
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Nice one 👍 , thank you @baronfac! Have a great day!

Ebazhanov:
@all-contributors please add @baronfac for design, code

PR #2350, opened 2021-10-06
New mysql questions

DavidsDvm:
- Added 4 new questions

  1. Q86. Why would you use a common table expression (CTE)?
  2. Q87. Which option modifier tells a program not to exit with an error if it does not recognize the option, but instead to issue a warning?
  3. Q88. You are working with the tables as shown in this diagram. You need to make sure that any record added to the purchases table consists of customerID, which already exists in the customer table, and a carID, which already exists in the cars table. You decide to use a trigger to do the validation. Which one do you use?
  4. Q87. What does this SQL statement return?
 
all the questions have answers, i check the SQL documentation to response that questions

Kindly merge the PR and mark it under Hacktoberfest label 2021. pls (;
Thanks & Regards.


Ebazhanov (approved):
Thank you @DavidsDvm for the new questions & answers and for your great contribution 👍 !

Ebazhanov:
@all-contributors please add @DavidsDvm for design, code

PR #2366, opened 2021-10-06
JSON: Add an undocumented question, fix Q2 answer

Nikostito:
Q2's answer was wrong as the answer given was not the closest JSON equivalent. The answer documented was an array and we need an object for JSON.


Ebazhanov:
@all-contributors please add @Nikostito for design, code

PR #2368, opened 2021-10-06
AWS: Add answers to Q72 and Q75

hectcastro:
- Add answer to Q72; fix Fargate typos
- AWS: Add answer to Q75

PR #2373, opened 2021-10-06
<#2368> Adding valid answers to Q72 and Q75 with reference link and also fixing typos for Q72

parikhparth23:
This PR contains valid answers for Q72 and Q75 with reference link and also fixing typos for Q72.
This issue was created by @hectcastro  - https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/2368


parikhparth23:
Removing this PR because of the fixed which is already there.

PR #2456, opened 2021-10-16
[MySQL] added q92

goodday451999:
added a new question in MySQL


monkrus:
@all-contributors please add @goodday451999 for data

PR #2459, opened 2021-10-16
Updated c++ quiz

jatinkumar027:


jatinkumar027:
Added question 98 in css quiz

PR #2476, opened 2021-10-17
Add Q101 in Microsoft Excel

Amsal1:


Amsal1:
> I think the correct answer is the 2nd

I am not sure myself, so added it without any answers 🙂

PR #2491, opened 2021-10-17
Add Q92 in MySQL

Amsal1:


monkrus:
@all-contributors please add @Amsal1 for doc

PR #2578, opened 2021-10-31
Fixed format of Q45 and Q58

diksha137:
Fixed format of Q45 and Q58 in PHP readme.


Ebazhanov (approved):
Thanks for improvement I hope prettier does not change it back! LGTM 👍

Issue #2626, opened 2021-11-13
Mongo DB Question 40 answer is Wrong


agrsachin81:
The correct answer is  "db.vehicle.distinct("category").length" not "db.vehicle.distinct("category").count()" count is not a known function of javascript array

PR #2627, opened 2021-11-13
#2626 length works with distinct

agrsachin81:
with find count() i.e. function works.
While with distinct length (variable) works


Ebazhanov (approved):
Good catch, thank you for your valuable contribution @agrsachin81 👍

Ebazhanov:
@all-contributors please add @agrsachin81 for design, code

PR #2684, opened 2021-11-25
Fixed variable names in Q9 for OOP quiz

masoudk1990:


masoudk1990:
#caeabc5: Inside the body of a static constructor, only static members of a class are accessible.

PR #2696, opened 2021-11-28
fix answer trojan / virus

thibault-ketterer:
viruses does not covers tracks
rootkit do, their goal is to be 100% transparent for the end user


monkrus (approved):

monkrus:
Great fix! @all-contributors please add @thibault-ketterer for maintenance

PR #2726, opened 2021-12-03
Update java-quiz.md

AAbozaid96:
Q51:
in the explanation: final member can also initialized inside instance-initializer block.


Ebazhanov (approved):
Thanks, LGTM! 💯

PR #2727, opened 2021-12-03
Update java-quiz.md

AAbozaid96:
Q100
Vactor Class extend Collection interface, so it's right option and the correct answer is HashMap class beacuse it implement Map interface


Ebazhanov (approved):
Looks good to me, thank you for the correction and reference link! 🥇

PR #2744, opened 2021-12-06
Fixed #2735

### LINKED ISSUE #2735: New questions in MS Power Point
**Why would you use Outline view?**
-  to quickly edit titles or bulleted text in the Slide pane
-  to view the comments on all slides
-  to add notes in the Slide pane
-  to change the font on all slides 

**Before publishing a presentation, which option should you use to scan a presentation and look for issues that may make it difficult for someone with a visual impairment to read?**
-  Protect Presentation
-  Inspect Document
-  Check Compatibility
-  Check Accessibility

**How can you filter the series and/or categories in a chart?**
- Edit the data to remove the data for the series or category.
-  Switch the rows and columns.
-  Use a filter so the data series or category does not display.
-  Change the chart type.

masoudk1990:


Ebazhanov (approved):
Thank you Masoud, great job! 👍

PR #2765, opened 2021-12-11
Update kotlin-quiz.md

ItSNeverLate:
class Cat (val name: String) {
  fun greet() { println("Hello ${name}") }
}

fun main() {
  val thunderCat = Cat("ThunderCat")
  thunderCat.greet()
}


Ebazhanov (approved):
Amazing 👍 , thanks!

PR #2786, opened 2021-12-11
Update kotlin-quiz.md

ItSNeverLate:
Q63 is similar to Q72
([reference](https://www.techiedelight.com/join-two-lists-kotlin/))
Q72 has some problems in value initiations


Ebazhanov (approved):
Thank you for the answers with reference! Great job 👍

PR #2789, opened 2021-12-11
Update kotlin-quiz.md

ItSNeverLate:
Wrong answer
and Wrong Question (Two correct answers)

"this" is redundand


Ebazhanov (approved):
you are unstop-able 👍

PR #2801, opened 2021-12-12
Update android-quiz.md

ItSNeverLate:
Adding Reference


Ebazhanov (approved):
Reference is everything, thanks for sharing you are great @ItSNeverLate !! :point_left:

PR #2808, opened 2021-12-12
Update android-quiz.md

ItSNeverLate:
Answer Correction


Ebazhanov (approved):
That was first class work. @ItSNeverLate

PR #2821, opened 2021-12-13
fix(linux): correct Q70 answer

edosrecki:
From https://man7.org/linux/man-pages/man1/bash.1.html:

> An additional binary operator, **=~**, is available, with the
              same precedence as == and !=.  When it is used, the string
              to the right of the operator is considered a **POSIX
              extended regular expression**


Ebazhanov (approved):
Outstanding! 👍 Thank you @edosrecki!

Ebazhanov:
@all-contributors please add @edosrecki for design, code

PR #2834, opened 2021-12-16
Fixed <h1> tag formatting

harmolipi:
The `<h1>` tag in this question was rendering as an actual html tag, causing it to not appear and throw off the formatting of the rest of the question. I just put it in backticks to resolve this.


Ebazhanov (approved):
Good catch 👍 , thanks for improvements!

PR #2865, opened 2021-12-26
Answer of Q40 will be scatter chart

gazihasanrahman:
Explanation: As both of the columns have numeric data, we will use a scatter chart and apply them in X-axis and Y-axis


Ebazhanov (approved):
Thanks for correction @gazihasanrahman 👍 , fyi I just added your explanation under the question.

PR #2915, opened 2022-01-07
Update python-quiz.md

griessej:
Update answer to question 112


Ebazhanov (approved):
Thank you @griessej for your pull request with example! 👍   It is seems to me a good idea to keep some examples, maybe later we can move them and collect in separate file.

Ebazhanov:
@all-contributors please add @griessej for design, code

PR #2963, opened 2022-01-21
Fixes Java Q100 answer

MRezaNasirloo:
HashSet makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time.

Reference: https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html

### DIFF
--- a/java/java-quiz.md
+++ b/java/java-quiz.md
@@ -1294,6 +1294,10 @@ class TaxCalculator {
 - [ ] contains unique elements
 - [x] contains sorted elements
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thanks for correction 👍 @MRezaNasirloo, I also have added your explanation and reference link fyi.

Ebazhanov:
@all-contributors please add @MRezaNasirloo for design, code

PR #2989, opened 2022-01-27
Fix the wrong answer for Ruby on Rails Q. 22

engr-hasanuzzaman:
The current answer for the Ruby on Rails Question 22 is not correct. The correct answer will be **Create an embedded Ruby file (.html.erb) and surround the Ruby code with `<%= %>`**. [Reference](https://guides.rubyonrails.org/action_view_overview.html#erb)


Ebazhanov (approved):
Thanks for the reference and correction @engr-hasanuzzaman great improvements! 👍

Ebazhanov:
@all-contributors please add @engr-hasanuzzaman for design, code

PR #2990, opened 2022-01-27
Fix the wrong answer for Ruby on Rails Q.38

engr-hasanuzzaman:
The current answer for the Ruby on Rails Question 38 is not correct. The correct answer will be `Product.where("name = ?", @keyword)` [Reference](https://guides.rubyonrails.org/active_record_querying.html#pure-string-conditions)


Ebazhanov (approved):
LGTM, thanks

PR #3026, opened 2022-01-30
Added a new question about dictionaries

Panoskontos:
I added this question because hash tables are  an important data structure in CS and there should be a question about how python implements them 

python_hash_table


Ebazhanov (approved):
Totally agree with you @Panoskontos and thank you for your valuable contribution! 👍

Ebazhanov:
@all-contributors please add @Panoskontos for design, code

PR #3038, opened 2022-01-31
Update css-quiz.md

SantiagoSiordia:
I answered some questions that appeared on my test, I will also add more questions to it. I got top 5% so I think they are correct.


Ebazhanov (approved):
Thanks for the answers and congratulation with the test! 👍

Ebazhanov:
@all-contributors please add @SantiagoSiordia for design, code

PR #3064, opened 2022-02-04
Q33 answer fixed in Go quiz questions

Imolai:
log.Error() is wrong answer, because it is not defined function in its package; see Explanation for the good answer.


Ebazhanov (approved):
Good catch 👍, thanks for detailed explanation @Imolai !

Ebazhanov:
@all-contributors please add @Imolai for design, code

PR #3068, opened 2022-02-04
Q30 and Q40 answer fixed in Go quiz questions

Imolai:
Fatal does not stop all tests, but marks the function as failed and stops its goroutine's execution. As Run runs f as a subtest in a separate goroutine, Fatal stops execution only of the subtest and continues with other test cases.

### DIFF
--- a/go/go-quiz.md
+++ b/go/go-quiz.md
@@ -411,10 +411,12 @@ Relevant excerpt from the article:
 
 #### Q33. How should you log an error (err)
 
-- [x] `log.Error(err)`
-- [ ] `log.Printf("error: %v", err)`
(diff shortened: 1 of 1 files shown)


Ebazhanov (changes requested):
@Imolai thanks again for detailed explanation, but I found that we have 2 similar question, could you please remove one them. Thank you in advance!

Imolai:
@Ebazhanov you are right, I removed the Q40 in commit 27dddf6827755b3ec3eec7d4e99e7635d4f4000f .

Ebazhanov (approved):
Thanks again! LGTM 👍

PR #3088, opened 2022-02-09
Correction in Spelling and Correction for question 22

ThaC0derDre:
Question #22 is either entered incorrectly, or the answer is incorrect.
To make the first choice correct, the question code should be written:
"var vals: Set<String> = ["4", "5", "6"]
vals.insert("5")"

Else, the correct answer is  "This code contains an error".


Ebazhanov (approved):
Thanks for correction @ThaC0derDre good catch! 👍

Ebazhanov:
@all-contributors please add @ThaC0derDre for design, code

PR #3160, opened 2022-02-20
question 95 - fix typo and add closing button tag

bganguly:
question 95 - fix typo and add closing button tag


Ebazhanov (approved):
Looks reasonable thank you @bganguly 👍 , I hope your fixes is looks the same as in original question!

PR #3209, opened 2022-03-04
React - Provided correct answer to question # 95 and fixed a typo

bganguly:
React - Provided correct answer to question # 95 and fixed a typo
- Existing line 1084 said 'kitteh'- which does not sound right - changed to 'kitten'
- the hug function just takes one parameter- 'name'- and throughout the react official tutorial on tic-tac-toe- there is no reliance on the explicit event object 'e'- so answers using 'e' are incorrect


Ebazhanov (approved):
Looks good, thanks for detailed explanation @bganguly 👍 , merged

PR #3210, opened 2022-03-04
Javascript - removed question # 123 which was duplicating # 82

bganguly:
Javascript - removed question # 123 which was duplicating # 82
Also updated the correct answer to 82.


Ebazhanov (approved):
Good catch 👍 , thank you @bganguly for your contribution!

PR #3306, opened 2022-03-18
Added 4 new questions and provided explanation.

0nyr:
* Q120. Which is a valid type for this lambda function?
* Q121. What is displayed when this code is compiled and executed?
* Q122. What variable type should be declared for capitalize?
* Q123. Which is the correct return type for the processFunction method?


Ebazhanov (approved):
Thanks a lot @0nyr for your really great job! 💯

PR #3352, opened 2022-03-25
C# fix Q44

eikowagenknecht:
Escape was missing in markdown + questionmark


Ebazhanov (approved):
Good catch, thank you @eikowagenknecht! 👍

PR #3370, opened 2022-03-26
docs: add new bash question

BigAl:
Add new bash question


Ebazhanov (approved):
Looks good, thanks! 👍

Issue #3343, opened 2022-03-24
CSS skill assment - some question marks


JMJHautier:
Hi there, 

in the CSS assessment,
 
 **Q43 ** is as follow: 
![image](https://user-images.githubusercontent.com/13255701/159917961-27a7e038-3120-4a88-8ebf-19b96adc9325.png)

The suggested answer would only select internal hyperlinks, sure, but only those that have no references. To select all internal hyperlinks of a page, one would need to select all href that _starts_ with "#", right? 

Hence, should'nt the right answer be: 
`   a[href^="#"]`
?

**Q112** has no answer indicated, but I guess it would be: 
![image](https://user-images.githubusercontent.com/13255701/159938478-d6e80ad9-02b6-433d-8ebb-d66278a34d47.png)

**Q119** also has no answers indicated, I guess it would be: 
![image](https://user-images.githubusercontent.com/13255701/159939589-5dccd1d5-a2d7-43fc-9763-4254b4a0c44a.png)

masoudk1990:
@JMJHautier What you said about Q43 and Q112 is correct. But the correct answer for Q119 is:
```
section + p {
  color: blue;
}
```
It means select the first `<p>` immediately after `<section>`

JMJHautier:
> @JMJHautier What you said about Q43 and Q112 is correct. But the correct answer for Q119 is:
> 
> ```
> section + p {
>   color: blue;
> }
> ```
> 
> It means select the first `<p>` immediately after `<section>`

Oh yeah, my bad. Thanks for the correction and related PR.

PR #3373, opened 2022-03-26
Fixed: #3343

masoudk1990:
Fixed issued in #3343 and also removed duplicated parts of Q112


Ebazhanov (approved):
LGTM, thank you @masoudk1990 🥇

PR #3374, opened 2022-03-26
docs: add 3 new aws-lambda questions

BigAl:
Add new aws-lambda questions. Not sure the answer they are looking fro Q47 as none of those match.


Ebazhanov (approved):
Ok, thanks 👍

PR #3378, opened 2022-03-26
[Linux] Add answers Q78 to Q81

### LINKED ISSUE #3229: Missing Linux assignment answers
image

pestadieu:
Fixes #3229 

Ping @D4RIO @Amsal1 
I am unsure about the answer to Q81. It would be nice if one of you could double check that.


D4RIO (changes requested):
Thank you for your collaboration, and for pinging us for review. The 81 should be the last option. See, fdisk -l will indeed read the contents of /proc/partitions if you don’t provide an argument, but to get partitions from there, and then it will read each table. The output of fdisk -l is a lot when compared to /proc/partitions.

its0x08 (approved):

Ebazhanov (approved):
Hello, @pestadieu! Thank you for your great input! 👍

Ebazhanov:
@all-contributors please add @pestadieu for design, code

Issue #3394, opened 2022-03-27
Rest API


Ebazhanov:
![image](https://user-images.githubusercontent.com/13170022/160298896-f2b5c4c9-d9ef-4202-afa2-6826c1a8ca62.png)

Issue #3389, opened 2022-03-26
Python


Ebazhanov:
![image](https://user-images.githubusercontent.com/13170022/160256459-b1109c69-6a36-4633-886b-dfaaa1680f4d.png)

rangaomkaram:
Option D : 
is the accurate implementation in python according to Set inbuilt methods.

Ebazhanov:
Hello, @rangaomkaram! if it is question, I do not know because I am not an expert. If you want we can double check with [@tik9](https://github.com/tik9), [@Amsal1](https://github.com/Amsal1) or create PR which we can merge. Thank you in advance!

rangaomkaram:
Option D: in the image is the correct answer for that question and here I am attaching the explanation to it 

#where take two sets
A = {1, 2, 3, 5, 7,65,56,8}
B = {1, 2, 4, 8, 9,45,12,65}
def jaccard_similarity(A, B):
    #Find intersection of two sets
    nominator = A.intersection(B)

    #Find union of two sets
    denominator = A.union(B)

    #Take the ratio of sizes
    similarity = len(nominator)/len(denominator)
    
    return similarity
similarity = jaccard_similarity(A, B)

print(similarity)

Ebazhanov:
please create PR, l like your explanation! ❤️

masoudk1990:
I didn't fix this issue. Why my PR is linked to this?

KAUTH:
I agree with @sjebbara that option B is the correct answer.
We want the ratio of the _size of the intersection_ to the _size of the union_ of the two sets. 

As reference, we can use [https://docs.python.org/3/tutorial/datastructures.html?highlight=set#sets]( https://docs.python.org/3/tutorial/datastructures.html?highlight=set#sets). 
>  a & b                              # letters in both a and b
>  a | b                              # letters in a or b or both

Ebazhanov:
thank you guys ❤️  I have added this question with answers and reference!

PR #3398, opened 2022-03-28
Fixed: #3394

masoudk1990:


rangaomkaram:
Python #3389 

Option D: in the image is the correct answer for that question and here I am attaching the explanation to it

#where take two sets
A = {1, 2, 3, 5, 7,65,56,8}
B = {1, 2, 4, 8, 9,45,12,65}
def jaccard_similarity(A, B):
#Find intersection of two sets
nominator = A.intersection(B)

#Find union of two sets
denominator = A.union(B)

#Take the ratio of sizes
similarity = len(nominator)/len(denominator)

return similarity
similarity = jaccard_similarity(A, B)

print(similarity)

PR #3477, opened 2022-04-09
Add correct answer and image to Python Q132

KAUTH:
Added _def jaccard(a, b): return len (a & b) / len (a | b)_ as the correct answer, as discussed in #3389.

Removed the link to https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/master/python/images/Q132.png, since it did not point to any existing resource yet and added the mathematical expression of the Jaccard Similarity written in LaTeX. Since the GitHub markdown does not support this natively, we add it as an image source from http://www.sciweavers.org/free-online-latex-equation-editor.


Ebazhanov (approved):
Looks good 👍 , thanks a lot for improvements @KAUTH

Ebazhanov:
@all-contributors please add @KAUTH for design, code

PR #3483, opened 2022-04-10
Update and add questions to Linux

goosepirate:
I'm not sure about the answer to Q87, I think it might be _They provide inheritance for user and group permissions_?


monkrus (approved):

monkrus:
@all-contributors please add @goosepirate for code

monkrus:
@all-contributors please add @goosepirate for example

PR #3488, opened 2022-04-10
Improved Q124

vahidmafi:
Added missing text-shadow class in <style>


Ebazhanov (approved):
Good catch 👍 , thank you @vahidmafi!

PR #3529, opened 2022-04-15
Update unity-quiz.md

Maselino:
Q24: According to my tests and understanding, the question asks in what relation to the variable "PlayerName" the pixels are inserted. As [Space(20)] is written above the variable PlayerName, it will also insert empty space above the variable in the Object inspector.


monkrus:
Thank you for the explanation, nice addition to the collection!

monkrus:
@all-contributors please add @Maselino for code

PR #3543, opened 2022-04-19
answer of Q41 and Q43

x24870:
Add the answer to Q41 with an example,
as well as the answer to Q43 with reference.


Ebazhanov (approved):
Thank you for the answers and references! 💪

PR #3553, opened 2022-04-20
T-SQL update. 7 new questions + formatting

BenVlodgi:
- Added 7 new T-SQL questions (3 answered).
- Update formatting in whole document.
- Update image with better quality version.
- Add another answer to question 32 because it had a different correct answer when I encountered it. Possibly previous answer was a typo.
- Remove wrong question text from question 33 (It had been copy-pasta'd from previous question).


Ebazhanov (approved):
Thank you @BenVlodgi, you did the great contribution! 👍 
Have a great day!

PR #3672, opened 2022-05-17
🛠️ Fixed Weird MarkDown Output

KhaledR57:
When a code block comes after a task list, it does not get parsed and appears like this ``` - [x] ``` so I added span tag after the task list to fix it.


Ebazhanov (approved):
Good catch 👍 , thank you @KhaledR57 for your improvements looks beautiful now!

Ebazhanov:
@all-contributors please add @KhaledR57 for design, code

PR #3785, opened 2022-06-03
Update Q10

hocandres:
The only difference is if you don’t specify the visibility (public, private or protected) of the members, they will be public in the struct and private in the class.

Q10. Which of the following is not a difference between a class and a struct?
- [ ] Because structs are part of the C programming language, there are some complexity between C and C++ structs. This is not the case with classes.
- [ ] Classes may have member functions; structs are private.
- [x] The default access specifier for members of struct is public, whereas for member of class, it is private.
- [ ] Template type parameters can be declared with classes, but not with the struct keyword.


Ebazhanov (approved):
Thank you @hocandres for your PR, i just added your explanation inside. Thanks again for your great input! 💪

Ebazhanov:
@all-contributors please add @hocandres for design, code

PR #3822, opened 2022-06-12
Fix Q.70 answer

devyuji:
That function does not need return keyword as it return value. The main issue there is React does not allow render more than one element.


Ebazhanov (approved):
Good catch, thanks a lot @devyuji 👍
if you want you can also add your explanation under the question, to let other people know it. Thanks

Ebazhanov:
@all-contributors please add @devyuji for design, code

PR #3873, opened 2022-06-19
New questions and answers in Python and Agile Methodologies

jgbispo:
I created some new questions that appeared to me and that I couldn't find in the repository and I also answered some questions that were unanswered


Ebazhanov (approved):
Looks perfect, thank you @jgbispo 💯

PR #3892, opened 2022-06-22
2 new questions

leighboz:
The browser doesn't seem to render checkboxes when the options don't have any text, it comes up as:

• []

Is this by design? Doesn't seem right to my eyes so I've added A, B, C, D text for 2 questions where that was happening.

Also there were 2 new questions I spotted on LinkedIn.


Ebazhanov:
@all-contributors please add @leighboz for design, code

PR #3977, opened 2022-07-09
Bug: Replace while with for

mabioIa:
Code sample on L105 was erroneously written using a while loop instead of a for loop


Ebazhanov (approved):
Thank you @mabioIa 👍, good catch!

Ebazhanov:
@all-contributors please add @mabioIa for design, code

Issue #3941, opened 2022-07-05
Microsoft Power BI


Ebazhanov:
![image](https://user-images.githubusercontent.com/13170022/177254473-44831985-da9b-4fb6-ab72-161f290bf242.png)
![image](https://user-images.githubusercontent.com/13170022/177254501-bb18ae89-4267-4ac0-ae7c-ab1231c08fba.png)

Just2Deep:
i have created a PR for this, can you please check

Ebazhanov:
@Just2Deep thanks for your contribution!

PR #3995, opened 2022-07-12
Microsoft Power BI #3941

Just2Deep:
Updated Power BI questions


Ebazhanov (approved):
Thank you @Just2Deep for the new question with reference links!!!! Looks perfect 🥇

Issue #4009, opened 2022-07-14
New question for Git


markleung0905:
![Screenshot 2022-07-14 at 21 00 50](https://user-images.githubusercontent.com/57431955/179072856-b32ca4e2-7457-4d14-85df-a673bdf671a1.png)

Ebazhanov:
Thank you guys @markleung0905, @Just2Deep i see the question is already in codebase! 🎉

PR #4020, opened 2022-07-16
New question for Git #4009

Just2Deep:
New question for git


Ebazhanov (approved):
Thank you @Just2Deep for the new questions, LGTM 👍

PR #4068, opened 2022-07-25
Request to update and change Q62

OskyEdz:
Hi!

I took this quiz and found that Q62 didn't sit quite right with me.
Found this github and am no proposing this change.

I hope you can understand my reasoning to these two changes.

1. I'd like to update the question so that it actually make sense.
2. I'd like to change the answer so it reflects today's view on UEFI

I'v documented both changes in two commits under my fork.
Disregard patch 1 and 2 reference. I thought I made two patches not 2 commits.

### DIFF
--- a/it-operations/it-operations-quiz.md
+++ b/it-operations/it-operations-quiz.md
@@ -500,8 +500,8 @@ Idea behind the answer: let's go for a quick restart to see whether the problem 
 #### Q62. Tania was told to address an issue in the UEFI. How is it accessed?
 
 - [ ] by using the System Configuration Utility (MSCONFIG)
-- [x] from the BIOS
-- [ ] from the computer's startup menu
+- [ ] from the BIOS
+- [x] from the computer's startup menu
 - [ ] on the network
 
 #### Q63. What is the technical reason why you must properly eject media?
--- a/it-operations/it-operations-quiz.md
+++ b/it-operations/it-operations-quiz.md
@@ -497,7 +497,7 @@ Idea behind the answer: let's go for a quick restart to see whether the problem 
 - [x] 254
 - [ ] 256
 
-#### Q62. Tania was told to address an issue in the UEFI. How is it accessed?
+#### Q62. Tania was told to change a setting in the UEFI. How is it accessed?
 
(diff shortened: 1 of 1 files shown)


101br03k:
i agree that the answer is wrong, but is the question different from the linkeding quiz or is it that you want to propose changes to the linkedin quiz. if you want to propose changes to the linkeding quiz i recomend contacting linkedin.

Ebazhanov (approved):
Thanks a lot for your updated and for your valuable contribution! It is completely fine to do like this. 💯

Ebazhanov (changes requested):
now i understand @OskyEdz, can you maybe put your description under the question? And keep original question to make it searchable?

OskyEdz (commented):
Q62. Tania was told to address an issue in the UEFI. How is it accessed?
Was changed to 
Q62. Tania was told to change a setting in the UEFI. How is it accessed?

And the answer changed from
"from the BIOS" to 
"from the computer's startup menu"

The phrase "to adress an issues in UEFI" could be interpreted as to alter something about the UEFI. For example, a employee at Asus found that in some cases your not able to adjust how much voltage the CPU can get, that employees has now been tasked with addressing this issue.

I changed the answer because you would not access the UEFI from the BIOS. Only in rare cases or manufacturing testing experimental stuff, does a system have both UEFI and BIOS.

To get to either UEFI or BIOS on a system you usually go via the computer's boot menu or press a certain key to boot straight into that mode.

OskyEdz:
I'm not sure, but did you have in mind some kind of comment in the MD file?

I'll try that when I'm on my computer tomorrow.

OskyEdz:
@Ebazhanov how do you like this comment?

Ebazhanov:
> @Ebazhanov how do you like this comment?

now looks good, I just got an idea can we have two question original one and one with your correction? Sorry for late reply
this way we can get read this huge comment and keep align with other questions. Or let me know I can change it by myself. Thanks

Ebazhanov (approved):
Thank you @OskyEdz for your valuable contribution! 👍

Ebazhanov:
@all-contributors please add @OskyEdz for design, code

PR #4104, opened 2022-07-29
Correction to 3 Git questions.

getinerler:
Corrected answers: 
33. After committing the branches, the best thing to do is to remove the merged branches. Last option doesn't seem to be a correct answer.
108. Using -p with log command results in showing the differences in each commit. It's easy to check.
117. What's being asked in the question is not removing the file but removing the changes made. Therefore: git checkout HEAD -- config.properties. I also corrected the misspelled word in the question. It was written as "confin.properties".


Ebazhanov (approved):
Thanks you for the correction @getinerler, great job! 👍 
also if you want I can add you to the column responsible person for this quiz.
have a great weekend!

Ebazhanov:
@all-contributors please add @getinerler for design, code

PR #4119, opened 2022-07-31
Correct Q&A count for Accounting and Agile Methodologies

jefferies917:
Accounting quiz didn't have Q31 so renamed 32 to 31 and 33 to 32 so it didn't skip a number anymore.

Agile methodologies had incorrect formatting on the last 2 questions, ### instead of ####.

I also reflected these changes in README.md with the correct question and answer number for these 2 quizzes.


Ebazhanov (approved):
LGTM 👍, thanks a lot @jefferies917

PR #4125, opened 2022-08-01
new question encountered

maximumn3rd:
I think the answer is right as 123abc but if '123abc' is meant as "the result is a string datatype" then 3rd is correct


Ebazhanov (approved):
I am not an expert, anyway thank you for your contribution @maximumn3rd 💯

PR #4127, opened 2022-08-01
Git quiz correction

jefferies917:
Corrected number of questions and answers in README.md

Q2 also had 2 answers previously because: The command is git remote add. The new added connection can be named origin or new. The only constraints, although it is not documented AFAIK, is that the connection name needs to be acceptable to git-check-ref-format, and it cannot be repeated.

While this makes sense as you can configure Git like this, out of the box the default will always be origin. It wouldn't make any sense for 'git remote add origin' to not be the intended answer.


Ebazhanov (approved):
Thanks for correction and detailed PR message, LGTM 👍

PR #4128, opened 2022-08-01
double question

maximumn3rd:
the question text was probably pasted twice


Ebazhanov (approved):
Good catch 👍, thank you @maximumn3rd

PR #4135, opened 2022-08-02
Answer Agile Q77 and Q93

jefferies917:
Corrected Q&A count for agile methodologies and deleted a duplicate question (102 & 89).

Q77 I think is pretty clear according to the manifesto. Q93 I thought was odd but by process of elimination input-output processing should be the correct answer.


Ebazhanov (approved):
Luke thanks a lot for correction and reference link! 💪

PR #4136, opened 2022-08-02
grammatical fix and add question

emanuele-em:


Ebazhanov (approved):
New question! Thank you @emanuele-em 🤝

PR #4145, opened 2022-08-02
Correct answer for C++ #9

F-Sossi:
My reading of the question is that they are looking for the false answer and while answer B is rather odd the one that is not true by my reading is D as you can use templates with both structures and classes as per the updated reference.


Ebazhanov (approved):
Thank you for your PR and explanation. I am not an expert of this languages, but glad that you put reference link and explanation! Looks good 👍

Ebazhanov:
@all-contributors please add @F-Sossi for design, code

PR #4191, opened 2022-08-09
Answered Q51 for Django

jefferies917:
For a Django and React project you would need to write cors middleware for the project to behave properly.


Ebazhanov (approved):
Wohoohoooo!!! Thanks for updating!!! Looks like you passed ✅ the quiz!

PR #4192, opened 2022-08-09
Added image in Que 23 in MS excel.md 

Aakashvani:
The image was absent, so I added it.


Ebazhanov (approved):
Thanks for improvements and adding image! Looks beautiful 🌹 👍

PR #4200, opened 2022-08-12
after-effects to es, fr and it

johnfelipe:


Ebazhanov:
@all-contributors please add @johnfelipe for content, tutorial, mentoring

PR #4220, opened 2022-08-14
Added reactjs spanish translation + some fixes in the english version

bandinopla:
Translated the reactjs quiz to spanish and during that process i found and fixes some issues in the english version. 
Also added myself in the creadits on the main readme file. 

In regards to the fixes on the english version i found some repeated questions so opted to change the question a bit to make it unique.  Some errors in the code snippeds too and lacking the language tag on them.


Ebazhanov (approved):
Thanks for initiative to be responsible, feel free to add for more topics 😉 
regarding translation i am a bit sceptical because manual translation can be different what linkedin will provide. Just FYI not sure that we have to do it with translator so the people will have hard time to match a search question.

bandinopla:
> ...regarding translation i am a bit sceptical because manual translation can be different what linkedin will provide. Just FYI not sure that we have to do it with translator so the people will have hard time to match a search question.

in some cases [like this one](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/commit/f835c01a5f4468ccd4788d49e85be6bd11877834#diff-849662fe725f1c9d25d269b6dd4fd0e08155903854a338765f001463e541d3edL560) i felt the question itself was wrong. It asks "when the Hello component will be displayed" but in the code it is not being rendered, just being assigned to a variable, and there's not even any code suggesting it will be indeed displayed... i mean one could asume, but the code doesn't show that. The correction i made helps the user focus on the original intended test of the question: to see if the reader understands the conditional (ternary) operator.

Ebazhanov:
that's can happens, we are all human. As a simple solution I would prefer to add new question with correct task. And keep old version at the same time to make it searchable/match with original linkedin.

Ebazhanov:
@all-contributors please add @bandinopla for design, code, maintenance, content

PR #4318, opened 2022-08-30
fix(html-quiz): remove repeated question

imgul:
Removed the repeated question for:
> Q76. Which choice is the best way to mark up this layout?


Ebazhanov (approved):
You are right, thank you @imgul for improvements and contribution! 💯

Ebazhanov:
@all-contributors please add @imgul for design, code

Issue #4268, opened 2022-08-21
Django New Question


Rushijaviya:
![1](https://user-images.githubusercontent.com/59735978/185779781-4e198ead-6aec-4b9f-b1cf-16a00b10f42b.jpeg)

prashant5025:
Hey, I like to contribute to this project would you like to assign me this issue?

Ebazhanov:
Thank you guys! 👍  Closing because it is already in codebase!

PR #4329, opened 2022-09-02
Django New Question #4268

Just2Deep:
Previous question is already answered. updated answer and link for the latest question(Q75)


Ebazhanov (approved):
LGTM, thank you @Just2Deep for updates 💯

PR #4340, opened 2022-09-04
Update bash-quiz.md

smashcoder:
- Highlighted the correct answer for better understanding for the user
- Q31 Both the answers, in text and image are arranged properly, now user will have no more confusion in the option. (Earlier option C in Image form and Option D in text form | Now both option in Image and Text are same)
- Q56 Options C is separated and clear for user to understand that it is not Option D, both are different.


Ebazhanov (approved):
Looks interesting, thank you for improvements!

smashcoder:
Thank You For Approving My Pull Request

PR #4343, opened 2022-09-04
5% up 

emanuele-em:
here you are a lot of questions, now Power Automate can be used to pass test


Ebazhanov (approved):
Yes, now we can win the price! Thanks a lot @emanuele-em for your great job!!! I really appreciate it! :pray:

PR #4344, opened 2022-09-04
add questions

emanuele-em:
some questions seems to be equals but only the answers are the same, my advice is to keep separate different quesitons with same answers


Ebazhanov (approved):
I absolutely agree with you 👍 , the idea is to keep all questions searchable.

PR #4350, opened 2022-09-05
Question 104 -> Explanation

InnocentAnyaele:
Modified question 104 explanation 

Maybe I'm getting it wrong but explanation for question 104, where it refers to value of the lists as (the letters) should be (the numbers) since the list contains the values (numbers) of the dictionary


Ebazhanov (approved):
I agree with you, thanks for correction @InnocentAnyaele 💪

Ebazhanov:
@all-contributors please add @InnocentAnyaele for design, code

PR #4366, opened 2022-09-08
Update c++quiz.md

smashcoder:
Highlighted All The Correct Option

### DIFF
--- a/c++/c++quiz.md
+++ b/c++/c++quiz.md
@@ -10,7 +10,7 @@ printf("%d", !b);
 
 - [ ] False
 - [ ] 0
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Well done @smashcoder but I am worry about project here https://github.com/linkedin-faq
they are using initial codebase to transfer it to their web and show for user. For example https://linkedin-faq.github.io/md2practice/

Maybe with this highlights we will destroy UI a little bit.

Ebazhanov:
or maybe highlights will show the answer which is not gut for user who is doing this first time fyi @smashcoder

Issue #4345, opened 2022-09-05
New Python Question


Westly:
![linkedin_new_python_question](https://user-images.githubusercontent.com/3219403/188339587-d8242b78-792a-404a-9075-37357bbd0ab7.png)

quantixdevelop:
![image](https://user-images.githubusercontent.com/100802375/190666948-ff761795-5d78-4668-8c89-163d2ea8a738.png)

quantixdevelop:
![image](https://user-images.githubusercontent.com/100802375/190667054-7e885260-3bea-4067-8c9a-af5023905f0f.png)
![image](https://user-images.githubusercontent.com/100802375/190667070-0b46b4ef-1449-42c9-9943-796cd2892bf7.png)
![image](https://user-images.githubusercontent.com/100802375/190667082-13e377db-dd5f-4808-89f6-59c42f8c62a1.png)

jdelucca:
This issue should now be closed, as both questions are now answered in the python file.

PR #4375, opened 2022-09-11
New Python Question #4345

Just2Deep:


Ebazhanov (approved):
LGTM, thank you for the new question and reference link @Just2Deep

PR #4420, opened 2022-09-20
Update gcp-quiz.md

GTekSD:
I found this 2 questions

### DIFF
--- a/google-cloud-platform/gcp-quiz.md
+++ b/google-cloud-platform/gcp-quiz.md
@@ -443,3 +443,18 @@ Explanation: Google does not store your keys on its servers and cannot access yo
 - [x] Cloud Storage and Big Query
 - [ ] Cloud Pub/Sub and Cloud Spanner
 - [ ] Cloud Pub/Sub and Big Query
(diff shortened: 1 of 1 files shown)


manikandan-artium:
The question number is wrong. it should be 57 and 58.

Ebazhanov:
> The question number is wrong. it should be 57 and 58.

Hello, @manikandan-artium! Thank you for your feedback, I am not an expert to judge but if you could provide reasonable explanation or reference link we can add them directly under the question. 

Regarding this PR i would like to merge it since @GTekSD bring new questions to this repo.

So please create your own PR with improvements and explanation. Thank you in advance!

Ebazhanov:
@all-contributors please add @GTekSD for design, code

GTekSD:
Okay, I am just contributing some questions because i didn't find those questions here.
and I am not sure it answers is correct or not but i got that skill badge. maybe it is correct. 
I just added few questions, please check there answers and update it. I will also try to find out answers with reason from my side..
Thank you everyone.

PR #4463, opened 2022-09-28
Excel - update Q23 answer

shuaaibbadran:
combo chart is the one used to display two different data series as a different series type within the same chart


Ebazhanov (approved):
Thank you @shuaaibbadran for your correction and valuable contribution! 👍

Ebazhanov:
@all-contributors please add @shuaaibbadran for design, code

PR #4479, opened 2022-09-29
Fixes #4475

### LINKED ISSUE #4475: Python Q54 (BST) Wrong Answer
Question asks for a True statement. The marked answer states:

> For any given Node in a binary Search Tree, the child node to the left is less than the value of the given node and the child node to its right is greater than the given node.

Which is incorrect. 

By [definition ](https://en.wikipedia.org/wiki/Binary_search_tree) each node should be greater than **all nodes in it's left subtree** and less than **all ones in its right subtree**.

As an example, the below tree meets the conditions stated in the answer, but it's not a BST because a node in the 8's left subtree is greater than it (11 is greater than 8).
```
       8
    /     \
   5      10
 /   \
3    11
```

mehhran:
Fixes the answer to the Python question (Q54) about Binary Search Trees.

### DIFF
--- a/python/python-quiz.md
+++ b/python/python-quiz.md
@@ -750,9 +750,9 @@ def __init__(self):
     pass
 ```
 
-#### Q54. Which of the following is TRUE About how numeric data would be organised in a binary Search tree?
+#### Q54. Which of the following is TRUE About how numeric data would be organised in a Binary Search Tree?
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thank you @mehhran for your improvements and valuable contribution! 💯

Ebazhanov:
@all-contributors please add @mehhran for design, code

Issue #4485, opened 2022-09-30
Python question: What does [X*2 for x in range(1,n)] return?


Bart-Source:
![New Python Question 1](https://user-images.githubusercontent.com/64977445/193269578-7cbc9449-232a-4c83-82b7-f962bf8a7a70.jpg)
I assume this returns a list with all even numbers **less** then 2*n since it starts at 1 and ends before n.
[https://www.w3schools.com/python/ref_func_range.asp](https://www.w3schools.com/python/ref_func_range.asp)

Just2Deep:
Yes, it returns all even numbers less than 2*n.

PR #4492, opened 2022-10-01
new question python #4485

Just2Deep:
Python question: What does [X*2 for x in range(1,n)] return? 

Added the question and reference for answer.


Ebazhanov (approved):
Thank you @Just2Deep, LGTM 👍

PR #4504, opened 2022-10-01
Added answer and reference to Angular Q55

nateonmission:
I've look over several articles (see my comment in discussion). I believe the answer is D "It configures the router to immediately load all routes that have a loadChildren property (routes that are typically loaded when requested)"

Because,
The default behavior is Load on Demand (as stated in the parenthetical on the response);
It does, in fact, load all the routes that have configured with the loadChildren statement;
It do not find anything indicating that instances of services are automatically created.


Ebazhanov:
@all-contributors please add @nateonmission for design, code

PR #4564, opened 2022-10-05
add new question number 139 css

eloygplaza:
To insert an external style sheet file into another style sheet file, which rule should you use?


Ebazhanov (changes requested):
Hello, @eloygplaza! Thanks for PR, but can you please look at different view 
image
I think we should improve it before, thanks

eloygplaza:
Hello @Ebazhanov,
I fixed the view thanks
Amazing repository :)

PR #4585, opened 2022-10-06
 #Hacktoberfest Fixed a wrong answer and added a reference

mhaendler:
Hey happy #Hacktoberfest, i stumbled across this repository and found something. 

The Q95 had a wrong answer, as the loop stops at 4 by break and not at 2. (also if you are really clear, the loop stops at 3 by break statement and afterwards increments $i by 1 so the output is 4. But it is what it is.

Also added a reference to Q1 because of the non-strict comparison between numbers and non-numeric strings. As there is a major difference between php version <8.0 and 8.0>.

I also started translating this questions into german, but this is going to take some time.

### DIFF
--- a/php/php-quiz.md
+++ b/php/php-quiz.md
@@ -11,6 +11,7 @@
 
 **_Both sides of the "spaceship" are equal, so the answer is 0. PHP will convert '76 trombones' to 76 in this context, as the string starts with '76'. Try it!_**
 **_For php 8.0 and forward the answer is [x] -1, for previous versions the answer is [x] 0._**
+[PHP 8 changed the way non-strict comparison between numbers and non-numeric strings work.](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.string-number-comparision)
 
 #### Q2. Which is the most secure way to avoid storing a password in clear text in database?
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thank you for correction and reference! 
I really appreciate your initiative about translation, but first I would suggest if Linkedin has this translation already, just turn your Languages to DE and see if the question will be in DE accordingly. 
Thanks @mhaendler your for contribution btw 💪

Ebazhanov:
@all-contributors please add @mhaendler for design, code

mhaendler:
> Thank you for correction and reference! I really appreciate your initiative about translation, but first I would suggest if Linkedin has this translation already, just turn your Languages to DE and see if the question will be in DE accordingly. Thanks @mhaendler your for contribution btw 💪

Alright done. I switched the language to "de" and the question's are accordingly then. So there is no translation needed for the questions? As for example in "c++" there is this file: "[c++quiz-de.md](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/main/c%2B%2B/c%2B%2Bquiz-de.md). Don't we need a similar file for PHP?

Ebazhanov:
@mhaendler it is up to you, any contribution are welcome. I can't control your free time! Have a great day und Ein schönes Wochenende!

PR #4596, opened 2022-10-06
Added a few questions

4Lajf:
Not 100% sure is what I answered was right so I'll leave it blank.


Ebazhanov (approved):
Looks perfect, no worries about answer! Thanks @4Lajf for contribution 👍

PR #4637, opened 2022-10-08
Python Quiz

Tushar-013:


Ebazhanov (changes requested):
@Tushar-013 Thanks for your PR, but could please check if those questions are already in https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/main/python/python-quiz.md

PR #4655, opened 2022-10-09
Corrected the answer to the question

nuromirzak:
There is an article on the official MongoDB website that explains JSON and JSON, and it says that `MongoDB stores data in BSON format both internally, and over the network, but that doesn’t mean you can’t think of MongoDB as a JSON database. Anything you can represent in JSON can be natively stored in MongoDB, and retrieved just as easily in JSON.`


Ebazhanov (approved):
LGTM, you can also put your explanation in the next PR.

PR #4665, opened 2022-10-09
Added new a question with answer

ShaliniChavan:
Added new a question with answer.
I have encountered.

Reference: 
https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById#:~:text=querySelector()%20and%20Document.,element%20objects%20in%20the%20DOM.

image


Ebazhanov (approved):
LGTM, thank you! 👍

Issue #4155, opened 2022-08-04
Accounting


Ebazhanov:
![image](https://user-images.githubusercontent.com/13170022/182773189-7f5fe562-0e44-4610-9a42-19d118b6b24c.png)
![image](https://user-images.githubusercontent.com/13170022/182773214-a5fc5d3e-b95e-442b-8dc2-4076b5adad84.png)
![image](https://user-images.githubusercontent.com/13170022/182773241-8f317a4d-060e-4016-a8a3-4ad03bf60ce8.png)
![image](https://user-images.githubusercontent.com/13170022/182773261-61d4cce8-9e73-422f-b824-7dd33e91a1df.png)
![image](https://user-images.githubusercontent.com/13170022/182773291-7280124c-2831-4d48-bcdb-838237d64ee9.png)
![image](https://user-images.githubusercontent.com/13170022/182773306-8c662746-3850-46ec-850d-e2c18b0c8d52.png)
![image](https://user-images.githubusercontent.com/13170022/182773327-50aa7b12-2547-42e3-bf3d-ba56aea14e67.png)
![image](https://user-images.githubusercontent.com/13170022/182773346-8d17b093-2ea0-47f2-baf9-88cfdbc612ce.png)

parikhparth23:
I am working on this

parikhparth23:
@Ebazhanov plz have a look at this https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/4733 and merge it

parikhparth23:
created a PR - https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/4733

PR #4733, opened 2022-10-12
[#4155] Added questions and its related answer with source

parikhparth23:
[#4155] Added questions and its related answer with source


Ebazhanov (approved):
Looks perfect 👍 , thank you @parikhparth23

PR #4763, opened 2022-10-14
Update php-quiz.md

GTekSD:
same question but


php/php-quiz.md
```diff
@@ -1284,7 +1284,7 @@ while ($fruit_name = current($array)) {
     }
 ```
 
-#### Q98. What does this code return?
+#### Q98. What does this code return? What does this code print?
```
Ebazhanov, line 1287: please make a duplication, this is not best practice. Later someone could remove `What does this code return?` let's keep both, thanks @GTekSD

Ebazhanov (changes requested):
left comment

php/php-quiz.md
GTekSD: #### Q98. What does this code return?
#### Q98. What does this code print?
Does it look okay? whenever someone searching for this question if "return" or "print" will show result not error sound. before I searched "What does this code return?" then its gives not found error, thats why I update for both. Thanks @Ebazhanov

Ebazhanov (approved):
LGTM 👍

PR #4868, opened 2022-10-18
Update java-quiz.md

eltohn:
What code is needed at line 8 ?


Ebazhanov (changes requested):
@eltohn could you please format: question and code under the question (you can see examples above), thanks

PR #4879, opened 2022-10-18
Update HTML quiz Q112

chandolakul:
Update HTML question 112
Please add this and add the label
hackoctoberfest2022-accept


html/html-quiz.md
```diff
@@ -1936,3 +1936,13 @@ A disabled fieldset is unusable and un-clickable. [Source](https://www.w3schools
...
+
+#### Q112. How to display preformatted text in HTML?
+
+- [ ] <p>
+
+- [x] <pre>
+
+- [ ] <hr>
+
+- [ ] All of the above
```
auchynnikau, line 1948: Let's format it a bit according to common style guide

```suggestion
- [ ] <p>
- [x] <pre>
- [ ] <hr>
- [ ] All of the above
```

auchynnikau (changes requested):

PR #4883, opened 2022-10-18
Javascript. Fix answer for question8. There was a wrong answer

artemu78:
Javascript. Answer to question 8 was not correct.  
There should be closure in answer.


javascript/javascript-quiz.md
```diff
@@ -121,7 +121,7 @@ for (var i = 1; i <= 4; i++) {
 }
 ```
 
-- [x] .
+- [ ] .
```
auchynnikau, line 124: Previous answer was correct you can check code results in browser console. There is required to find code which will log sequence of numbers from 1 to 4.

Answer selected before:
image

Your answer:
image

Ebazhanov:
@artemu78 thanks, based on the comment closing.

javascript/javascript-quiz.md
artemu78: you right, thanks for check
i missed checkbox location, so checked checkbox stay BEFORE its answer, i missed this

PR #4930, opened 2022-10-19
Add html-quiz

isha-dhek:


html/html-quiz.md
```diff
@@ -1950,3 +1950,10 @@ A disabled fieldset is unusable and un-clickable. [Source](https://www.w3schools
 - [ ] Specifies a context menu for an element. The menu appears when a user right-clicks on the element.
 - [x] Specify whether the content of an element should be editable or not
 - [ ] Return the position of the first found occurrence of content inside a string
+
+#### Q113. What is the correct HTML for inserting a background image?
```
artemu78, line 1954: Good question, but may be it is more about CSS.  
Btw we have such qiestion in CSS (question 118) https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/blob/main/css/css-quiz.md#q118-which-style-rule-would-you-apply-to-set-the-background-image-to-display-the-contents-of-the-woodpng-file

Ebazhanov: @isha-dhek could you please double check, thanks

PR #4994, opened 2022-10-22
Create m Coloring Problem.py

code-with-utkarsh:
m Coloring Problem using Backtracking in Python


Ebazhanov:
@code-with-utkarsh closing since it is not relevant to any of the quizes

PR #5020, opened 2022-10-24
Multiline comments in c# is wrong Question18

pedrolopes101:
Not sure if all answers should be wrong or the person who wrote the answers maybe made a mistake. Updated to what it should be. You can even check by clicking on the https://www.w3schools.com/cs/cs_comments.php that is linked below the question

### DIFF
--- a/c-sharp/c-sharp-quiz.md
+++ b/c-sharp/c-sharp-quiz.md
@@ -220,7 +220,7 @@ foreach(string employee in employeeQuery)
 - [ ] //\* Multiline
       / Single Line
 - [x] // Single Line
-      /_ Multiline _/
+      /* Multiline */
 
 [w3schools: C# Comments](https://www.w3schools.com/cs/cs_comments.php)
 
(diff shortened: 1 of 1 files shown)


c-sharp/c-sharp-quiz.md
```diff
@@ -220,7 +220,7 @@ foreach(string employee in employeeQuery)
 - [ ] //\* Multiline
       / Single Line
 - [x] // Single Line
-      /_ Multiline _/
+      /* Multiline */
```
Ebazhanov, line 223: it could be auto formatted with prettier
can you please also add answers inside backtick "```" to freeze the answers, thanks

Ebazhanov (changes requested):
Thanks for your PR @pedrolopes101, left comment

c-sharp/c-sharp-quiz.md
pedrolopes101: Did this in my new commit can you please check?

Ebazhanov:
@all-contributors please add @pedrolopes101 for design, code

PR #5041, opened 2022-10-26
Update object-oriented-programming-quiz.md

shabzkhan:
Which relationship best illustrates an abstract-concrete class relationship?

this question has one correct and other wrong answer, I removed incorrect one


Ebazhanov (approved):
LGTM, thank you @shabzkhan 💪

PR #5071, opened 2022-10-30
Update Question Formats

mustafapoya:
updated the questions texts, styles, codes, tags in the same format


Ebazhanov (approved):
Thank you @mustafapoya taking care for all of this! 🍻

PR #5073, opened 2022-10-31
Updated question formats

mustafapoya:


Ebazhanov (changes requested):
@mustafapoya please double check, there is no any changes in your PR

PR #5074, opened 2022-10-31
Periodic table

Shivampatel07:
This is a program which uses hashing with linked list to print a chemistry elements.


Ebazhanov:
thanks, but this wrong repo your programm

Issue #3946, opened 2022-07-06
Is the Answer to React Q19 correct?


aderchox:
Q19. When do you use useLayoutEffect?
[ ] to optimize for all devices
[ ] to complete the update
[ ] to change the layout of the screen
[x] when you need the browser to paint before the effect runs

Although the fourth option is checked, this is what I've found on web about useLayoutEffect which is completely against the checked option:
> Updates scheduled inside useLayoutEffect will be flushed synchronously before the browser has a chance to paint.

From: [LogRocket](https://blog.logrocket.com/useeffect-vs-uselayouteffect-examples)

jefferies917:
I believe this question is in fact correct. On reading the docs [here](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) it looks like useLayoutEffect stops updates occurring until after the page has loaded so that the effects run after the user can see the page. Otherwise seeing all the effects while elements are loading in would be bad UX.

NaNshekhar04:
Hello @aderchox the answer Looks correct ! As the **useLayoutEffect** function is triggered synchronously before the DOM mutations are painted. It will run after every **_render_** but _**before the screen is updated.**_ . Looks at the Example Below this will give you a fair good idea on how it works ! 

I would Like to refer you to go through this article on **useLayoutEffect** all of your doubts will be cleared up.  👍

Ref :- (https://react-hooks-cheatsheet.com/uselayoutEffect)


![ans](https://user-images.githubusercontent.com/110991877/197322276-9574e2c0-af2c-413e-9901-595f72431f2f.gif)

IMPranshu:
Yes, I think the answer is correct. Because the rest of the options are factually not possible by the useLayoutEffect

dagerzuga:
> You’ll notice that the useLayoutEffect gets executed before the useEffect hook without much concern for DOM mutation.

SRC: https://www.knowledgehut.com/blog/web-development/react-uselayouteffect

Hey @aderchox the question itself can be a bit confusing, but yes, as the other folks commented, the answer is correct.

@Ebazhanov This issue can be closed. [This PR](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/5123) expands that answer.

PR #5123, opened 2022-11-12
Expand answer of Q19 - React.js

dagerzuga:
## Description
Adding an explanation for question 19 of the React.js assessment after seeing what was discussed in this issue: https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/issues/3946. That same issue is also being linked in the question as a future reference.


Ebazhanov (approved):
LGTM, thanks a lot @dagerzuga! 💪

Issue #5213, opened 2022-12-10
Missing question Python


edunzer:
I'm submitting an ISSUE: please check one with "x"

- [x] New question{'s} 
![Picture2](https://user-images.githubusercontent.com/20893684/206876157-a4d264d3-f28e-46e9-b25d-0c89b935ee96.jpg)

- [ ] New assessment 
- [ ] Missed questions/answers 
- [ ] Request for new quiz/answers/...
- [ ] Documentation issue or request for ...
- [ ] ...

dylannalex:
I'm working on it!

ed3899:
closed!

PR #5229, opened 2022-12-13
Add new Python question #5213

dylannalex:
## PR Checklist

### DOD

- [ ] I have made small correction/improvements

### Changes / Instructions

Here's the new question I added:

#### Q175. Suppose you have a class named MyClass which has multiple inheritance and methods with the same name in its ancestors. Which class method could you call to see which method will get priority when invoked?

- [x] `MyClass.__mro__`
- [ ] `MyClass.hierarchy()`
- [ ] `callable(MyClass)`
- [ ] `dir(MyClass)`

#### Explanation

Method Resolution Order (MRO) is the order in which Python looks for a method in a hierarchy of classes. It returns a list of types the class is derived from, in the order they are searched for methods.

#### Example

```python
class A:
    def __init__(self):
        pass

    def method(self):
        print("A.method() run")


class B:
    def __init__(self):
        pass

    def method(self):
        print("B.method() run")


class C(B, A):
    def __init__(self):
        pass


c = C()
C.__mro__
>>> (<class '__main__.C'>, <class '__main__.B'>, <class '__main__.A'>, <class 'object'>)
```
This means that if we run `c.method()`, the instance `c` will call the method defined at class `B`:
```python
c.method()
>>> B.method() run
```
See more info [HERE](https://stackoverflow.com/questions/2010692/what-does-mro-do).

### DIFF
--- a/python/python-quiz.md
+++ b/python/python-quiz.md
@@ -2387,3 +2387,5 @@ my_list = [2, 'apple', 3.5]
 - [ ] `MyClass.hierarchy()`
 - [ ] `callable(MyClass)`
 - [ ] `dir(MyClass)`
+
+**Explanation:** MRO stands for Method Resolution Order. It returns a list of types the class is derived from, in the order they are searched for methods.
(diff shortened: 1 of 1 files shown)


Ebazhanov:
@all-contributors please add @dylannalex for design, code

PR #5256, opened 2022-12-22
Updated a wrong answer

unikdahal:
I updated the wrong answer which was present

PR #5219, opened 2022-12-11
44th answer updated

golamrabbiazad:
`~x` is doing a Unary complement (bit inversion).
Bitwise operators modify variables considering the bit patterns representing the values. So, If I simplify this as (-2 || false) then it obviously executes `Part A`.

## PR Checklist

This PR is ready for review and meets the requirements set out
in [Suggestion how to contribute](CONTRIBUTING.md)

### DOD

- [ ] I have added new quiz{'s}
- [ ] I have added new reference link{'s}
- [x] I have made small corrections/improvements

### Changes / Instructions

_Add instructions to me, please type here, thanks_

PR #5306, opened 2023-01-06
Fix wrong C++ answer(?s)

dehre:
## PR Checklist

## DOD

### Question 28: What is the value of `x` after running this code?
It seems to me that either the question or the answer is wrong.
Compiling and running this snippet with both `clang` and `gcc` prints `-3`:
```cpp
#include <iostream>
int main()
{
    int x = 10, a = -3;
    x = +a;
    std::cout << x << '\n';
    return 0;
}
```

The original answer, `-7`, would have been correct given this snippet instead:
```cpp
int x = 10, a = -3;
x += a;
```

@unikdahal I've seen your [recent PR here](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/5256), what do you think? Which compiler did you use?

### Question 44: What is the result from executing this code snippet?
```cpp
bool x=true, y=false;

if (~x || y) {
    /*part A*/
} else {
    /*part B*/
}
```

Original answer: Part A executes because the expression (~x || y) always results in true if y==false.

Well, it's correct that part A executes, but in my opinion the correct answer is: Part A executes because `~x` is not zero, meaning true.

Assuming the boolean value is represented using one byte, `true` becomes `00000001`, and flipping its bits you get `11111110`, which is a non-zero value.
Therefore the if-condition will always evaluate to true, no matter the value of `y`.
To confirm this theory, if you compile using the `-Wbool-operation` flag, with `clang` you get a warning stating: "bitwise negation of a boolean expression always evaluates to 'true'".

@golamrabbiazad you recently [updated this answer](https://github.com/Ebazhanov/linkedin-skill-assessments-quizzes/pull/5219), what do you think?

### Question 34: What is the output of this code?
```cpp
#include <cstdio>
using namespace std;

int main(){
    char c = 255;
    if(c>10)
        printf("c = %i, which is greater than 10", c);
    else
        printf("c = %i, which is less than 10", c);
    return 0;
}
```

Possible answers:
- [x] c = -1, which is less than 10
- [ ] c = 255, which is greater than 10
- [ ] c = -1, which is greater than 10
- [ ] c = 255, which is less than 10

The result is correct on my machine too, but, according to the standard, `char` could be either `signed` or `unsigned`; in the latter case, the answer would have been: "c = -1, which is greater than 10".
From the [reference](https://en.cppreference.com/w/cpp/language/types):
> The signedness of char depends on the compiler and the target platform: the defaults for ARM and PowerPC are typically unsigned, the defaults for x86 and x64 are typically signed.

So, either I checked something wrong, or LinkedIn should update its question.
Perhaps some veteran C++ developer can provide feedback?

### DIFF
--- a/c++/c++-quiz.md
+++ b/c++/c++-quiz.md
@@ -554,7 +554,8 @@ int main(){
 - [ ] c = -1, which is greater than 10
 - [ ] c = 255, which is less than 10
 
-Technically, a `char` could be either `signed` or `unsigned`; in the latter case, the second answer would be correct.
+Technically, whether a `char` is `signed` or `unsigned` is implementation-defined;
+in the latter case, the second answer would be correct.
 [Reference](https://en.cppreference.com/w/cpp/language/types)
 
 #### Q35. How can C++ code call a C function?
--- a/c++/c++-quiz.md
+++ b/c++/c++-quiz.md
@@ -417,8 +417,8 @@ x=+a;
 ```
 
 - [ ] 3
-- [x] 7
-- [ ] -3
+- [ ] 7
+- [x] -3
 - [ ] 13
 
 #### Q29. Which statement is true?
@@ -553,6 +553,9 @@ int main(){
 - [ ] c = 255, which is greater than 10
 - [ ] c = -1, which is greater than 10
 - [ ] c = 255, which is less than 10
+
+Technically, a `char` could be either `signed` or `unsigned`; in the latter case, the second answer would be correct.
+[Reference](https://en.cppreference.com/w/cpp/language/types)
 
 #### Q35. How can C++ code call a C function?
 
@@ -819,9 +822,9 @@ if (~x || y) {
 }
 ```
 
-- [x] Part A executes because the expression `(~x || y)` always results in true if `y==false`.
+- [ ] Part A executes because the expression `(~x || y)` always results in true if `y==false`.
 - [ ] Part B executes because the statement `(~x || y)` is invalid, thus false.
-- [ ] Part A executes because `~x` is not zero, meaning true.
+- [x] Part A executes because `~x` is not zero, meaning true.
 - [ ] Part B executes because `~x` is false and `y` is false, thus the `OR` operation evaluates as false.
 
 #### Q45. What would be the output of this code?


golamrabbiazad:
@dehre On the 44th question, You are absolutely right and I agree with that explanation.

Actually, the first time I thought that answer would be either `A` or `C` and I don't have a strong explanation for bit shifting to mark `C` as the answer. I go with the flow and pretty understandable answer `A` and I looked at the y value given as a false. 
**If the left part always returns true then doesn't matter if the right part is true or false which is super understandable.**

So, the right answer should be `C`. :+1: 

Please, add this explanation as a reference to this question. 

Reviewed, PR ready to Merge. :heavy_check_mark:

golamrabbiazad:
Alright, coming to the 34th question.

Yes, the `char` type could be `signed` or `unsigned`. 

> the defaults for ARM and PowerPC are typically unsigned

Yes, there are some compilers on some platforms that may define the char type as unsigned by default in order to optimize code for certain types of applications. 

It will be a long conversation if we go with that statement.

By default, the `char` type is `signed` on most systems. The `char` type is equivalent to the `signed char` type so, the c variable will be treated as a signed integer type, and it stored a negative value as -1 because of its range.

if it is unsigned then it would have been explicitly provided using `unsigned`.

I Hope, I clarified this.

dehre:
Exactly, `-1` is the correct answer *most of the times*.
Question is, should we make it clear, or just leave it out?
On the one hand, it would make the answer (a bit) more correct, or informative at least.
On the other, this document isn't a tutorial, there are barely some references here and there.

Ebazhanov:
> Alright, coming to the 34th question.
> 
> Yes, the `char` type could be `signed` or `unsigned`.
> 
> > the defaults for ARM and PowerPC are typically unsigned
> 
> Yes, there are some compilers on some platforms that may define the char type as unsigned by default in order to optimize code for certain types of applications.
> 
> It will be a long conversation if we go with that statement.
> 
> By default, the `char` type is `signed` on most systems. The `char` type is equivalent to the `signed char` type so, the c variable will be treated as a signed integer type, and it stored a negative value as -1 because of its range.
> 
> if it is unsigned then it would have been explicitly provided using `unsigned char`.
> 
> I Hope, I clarified this.

thanks for your comments @golamrabbiazad I have added you to the table in front of C++.

Ebazhanov (approved):
Thanks a lot @dehre for your valuable contribution! Great job 👍

Ebazhanov:
@all-contributors please add @dehre for design, code

golamrabbiazad:
> Question is, should we make it clear, or just leave it out? 

I think, what we discussed, is enough for the clearness. It's not that deep we are thinking about, it's a LinkedIn test.

Nice to have a conversation with you, thanks!

> On the one hand, it would make the answer (a bit) more correct, or informative at least. 

Yes, I wish :100:

PR #5386, opened 2023-01-21
Add references to Q51-60 for the Swift quiz

mediumhotcoffee:
## PR Checklist

This PR is ready for review and meets the requirements set out in [Suggestion how to contribute](CONTRIBUTING.md)

### DOD

### Changes / Instructions

Add references to questions 51-60 of the Swift quiz. Official docs only.

Improve the way the answers are written for consistency (and trying to match the way they are written on the actual quiz).

Note: where the name of the chapter/section from the docs would give the answer away, the name was omitted from the reference.

Formatted and indented code where necessary.

Fix the answer for question 56. Deinitialization in Swift takes place "immediately before a class instance is deallocated" (from the Swift book). So "When a class instance is being removed from memory" seems to be the correct answer.

"When a class instance needs memory" is a weird answer: "needs memory" is very vague. Let's say we instantiate a class that has a single Array property set to `nil`. If we assign an array to the property, it can be said that the class "needs memory" to store the array. A deinitializer wouldn't be called in that situation. In the iOS context, `UIViewControllers` have a `didReceiveMemoryWarning()` method. I guess it's trying to trick people in a weird way to think about `deinit` on ViewController classes? As I said, weird.

"When the executable code is finished" is also vague and weird. Which code? The app's? That's all executable code, so that's a pretty absurd proposition. The class'? It's possible for a class instance to be held by an optional variable that's set to `nil` while a strong reference prevents deallocation and, therefore, preventing the triggering of the deinitializer (see the ARC chapter in the Swift book for a simple example).

### DIFF
--- a/swift/swift-quiz.md
+++ b/swift/swift-quiz.md
@@ -597,7 +597,7 @@ x = 10.0
 #### Q50. What will this code print to the console?
 
 ```swift
-var items = ["a":1, "b":2, "c":"test"] as [String: Any]
+var items = ["a": 1, "b": 2, "c": "test"] as [String: Any]
 items["c"] = nil
 print(items["c"] as Any)
 ```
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Thank you for your great job! 👍

PR #5757, opened 2023-05-19
Add questions to C++ and Go quiz

DyingDown:
PR Checklist

### DOD

- [ ] I have made small correction/improvements

##### Q99. Once you are done writing to a file, what method will you call on the `ofstream` to notify the operating system?

- [ ] printout()
- [x] close()
- [ ] destroy()
- [ ] flush()

##### Q68. What value does a string variable hold if it has been allocated but not assigned?

- [ ] nil
- [ ] undefined
- [ ] null
- [x] ""

> If a string variable has been allocated but not assigned a value, its default value is an empty string "". In Go, uninitialized string variables are automatically assigned the zero value for their respective type, which for strings is an empty string.

##### Q69. Which built-in function is used to stop a program from continuing?

- [x] panic
- [ ] There is no such function.
- [ ] raiseException
- [ ] exit

> The built-in function used to stop a program from continuing is `panic()`. When `panic()` is called, it triggers a panic, which stops the normal execution flow of the program and begins panicking. If the panic is not recovered, the program terminates.

##### Q70. What will the output be?

```go
a,b := 1, 2
b,c:= 3, 4
fmt.Println(a, b, c)
```

- [x] 1 3 4
- [ ] 1 2 3
- [ ] 1 2 4
- [ ] It will not compile.

[Go Playground Example](https://go.dev/play/p/qUI50GNGWTq)

### DIFF
--- a/c++/c++-quiz.md
+++ b/c++/c++-quiz.md
@@ -1556,3 +1556,10 @@ X+=a;
 - [ ] 3
 
 **Explanation :** `+=` means increasing value. So `x += a` is equivalent to `x = x + a`
+
+#### Q99. Once you are done writing to a file, what method will you call on the `ofstream` to notify the operating system?
+
+- [ ] printout()
+- [x] close()
+- [ ] destroy()
+- [ ] flush()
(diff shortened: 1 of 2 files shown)


Ebazhanov (approved):
Thank you @DyingDown for adding new questions with answers! 👍

Ebazhanov:
@all-contributors please add @DyingDown for design, code

PR #5834, opened 2023-06-11
This fixes #5752 

### LINKED ISSUE #5752: Fix Answers in MongoDB assessment
I'm submitting an ISSUE: please check one with "x"
- Missed questions/answers 

### Q.24 A compound index allows you to _ ?
- Use more than one field per index.

Reference: https://www.mongodb.com/docs/v6.0/core/index-compound/
Argument regarding the wrong answer: https://www.mongodb.com/docs/v6.0/reference/collation/#operations-that-support-collation

TLDR: 

> To use an index for string comparisons, an operation must also specify the same collation.
> 
> You cannot specify multiple collations for an operation. For example, you cannot specify different collations per field, or if performing a find with a sort, you cannot use one collation for the find and another for the sort.
> 
> For a compound index where the index prefix keys are not strings, arrays, and embedded documents, an operation that specifies a different collation can still use the index to support comparisons on the index prefix keys.
### Q.45 What is NOT a standard role in MongoDB?
- delete collections

Reference: https://www.mongodb.com/docs/v6.0/reference/built-in-roles/#backup-and-restoration-roles
TLDR: You won't find "delete collections" role in the document above

### Q46. Which MongoDB shell command deletes a single document?
Answer is correct, just a note for future reference:

Note: db.collection.remove() is deprecated in the new mongosh. Use db.collection.deleteOne() or db.collection.deleteMany().
References:
- https://www.mongodb.com/docs/mongodb-shell/crud/delete/
- https://www.mongodb.com/docs/v6.0/reference/method/db.collection.remove/

### Q49. You need to get the names of all the indexes on your current collection. What is the best way to accomplish this?
Correct answer, just a note for future reference:

Note: An alternative method in the mongosh shell is listIndexes()

Reference: https://www.mongodb.com/docs/v6.0/reference/command/listIndexes/#mongodb-dbcommand-dbcmd.listIndexes

### Q50. You are going to do a series of updates to multiple records. You find setting the multi option of the update() command too tiresome. What should you do instead?
Correct answer, just a note for future reference:

Note: An alternative method for db is .update()

Reference: https://www.mongodb.com/docs/v6.0/reference/command/update/#mongodb-dbcommand-dbcmd.update

### Q52. Given a customer collection which includes fields for gender and city, which aggregate pipeline shows the number of female customers in each city? (Alternative: How can you view the execution performance statistics for a query?)
Correct answer, just a note for future reference:

Note: If you want to analyze the performance of a query use .explain("executionStats")
Reference: https://www.mongodb.com/docs/v6.0/tutorial/analyze-query-plan/

### Q55. Which file in the MongoDB directly holds the MongoDB daemon?
- mongod
Assuming that by file means executable. [Docs](https://www.mongodb.com/docs/v6.0/reference/program/mongod/)

### Q57. What is the most accurate statement regarding MongoDB and ad hoc queries?
- MongoDB allows ad hoc queries.
Note: You don't need an index to perform ad hoc queries. Only pick one choice

Reference: https://www.mongodb.com/docs/v6.3/query-api/

### Q.61 How can you improve the appearance of the output JSON that contains the _id?
This one is unclear.

1. db.collection.set() -> not a function
2. db.collection.format() -> not a function
3. Use $_id = value -> Assuming you do the following
 `db.sales.find( { _id: ObjectId("64653c284df505e68d65733e") })` then it is possible to look for an item based on the id. But again...the output remains the same
4. Creating a second index, does nothing to the output...it only means faster read times but slower write times.

### Q62. What happens to a Replica set oplog if it runs out of memory?
The oplog is capped collection and can't run out of memory

Argument:

Why "The oplog will be saved on one of the secondary servers." is wrong:
> MongoDB applies database operations on the [primary](https://www.mongodb.com/docs/v6.0/reference/glossary/#std-term-primary) and then records the operations on the primary's oplog. The [secondary](https://www.mongodb.com/docs/v6.0/reference/glossary/#std-term-secondary) members then copy and apply these operations in an asynchronous process. All replica set members contain a copy of the oplog, in the [local.oplog.rs](https://www.mongodb.com/docs/v6.0/reference/local-database/#mongodb-data-local.oplog.rs) collection, which allows them to maintain the current state of the database.

Reasoning behind the right answer:

> The [oplog](https://www.mongodb.com/docs/v6.0/reference/glossary/#std-term-oplog) (operations log) is a special [capped collection](https://www.mongodb.com/docs/v6.0/reference/glossary/#std-term-capped-collection) that keeps a rolling record of all operations that modify the data stored in your databases.
> 
> Unlike other capped collections, the oplog can grow past its configured size limit to avoid deleting the [majority commit point.](https://www.mongodb.com/docs/v6.0/reference/command/replSetGetStatus/#mongodb-data-replSetGetStatus.optimes.lastCommittedOpTime)

### Q63. MongoDB ships with a variety of files. Which file runs the MongoDB shell?
The answer is correct, just a note for future reference:

Note: `mongosh` is the new mongo shell, `mongo` is deprecated.

> Starting in MongoDB v5.0, 
[mongosh](https://www.mongodb.com/docs/mongodb-shell/)
 replaces [mongo](https://www.mongodb.com/docs/v6.0/reference/mongo/#mongodb-binary-bin.mongo) as the preferred shell.

Reference: https://www.mongodb.com/docs/mongodb-shell/

### Q67. Suppose you are using the mongoimport command to import personnel data and there is a unique index on the email field. What happens when there are duplicate emails in the import?

May be two possible right answers here:

- The command doesn't prompt you to correct the bad error, but logs an error and aborts the operation.
- It adds everything up to and not including the duplicate record with:

`E11000 duplicate key error collection: test.employees index: email_1 dup key: { email: "another1@example" }`

I wasn't able to experiment with mongoimport because of software limitations. The error I got above was from doing inserts with the `db.collections.insertMany` api

### Q69. You need to be able to quickly find a word in a text field. What should you do?

- Create a text index on the field and do a $text Query.

Argument: You need a text index in order to perform a $text query on a field. $text query uses the text index under the hood
References: 

- [$text query](https://www.mongodb.com/docs/v6.0/reference/operator/query/text/#examples)
- [Text index](https://www.mongodb.com/docs/v6.0/core/index-text/)

### Q71. After installing MongoDB on your machine, what must you do before launching Mongo?
- Create a data directory.

Note: The question in case is ambiguous. In the mongo docs, on the Windows Installation section, it clearly specifies the need for creating a [data directory](https://www.mongodb.com/docs/v6.0/tutorial/install-mongodb-on-windows/#run-mongodb-community-edition-as-a-windows-service). However, that does not seem to be the case for Unix based systems. So, that gives use the closest possible solution for a specific platform. If we extrapolate that to the question, that seems to be the most reasonable solution.

### Q79. You need to add an index to the large name collection in your production database. You do not want to have disruption of service for your users and you can't afford to have a team to do the work during after hours. What should you do?
Answer is correct, just a note for future reference: 

> Concurrency
> Changed in version 4.2.
> 
> MongoDB uses an optimized build process that obtains and holds an exclusive lock on the specified collection at the start and end of the index build. All subsequent operations on the collection must wait until 
> [createIndex()](https://www.mongodb.com/docs/v6.0/reference/method/db.collection.createIndex/#mongodb-method-db.collection.createIndex)
>  releases the exclusive lock. 
> [createIndex()](https://www.mongodb.com/docs/v6.0/reference/method/db.collection.createIndex/#mongodb-method-db.collection.createIndex)
>  allows interleaving read and write operations during the majority of the index build.
> 
> For more information on the locking behavior of 
> [createIndex()](https://www.mongodb.com/docs/v6.0/reference/method/db.collection.createIndex/#mongodb-method-db.collection.createIndex)
> , see [Index Builds on Populated Collections.](https://www.mongodb.com/docs/v6.0/core/index-creation/#std-label-index-operations)

Reference: https://www.mongodb.com/docs/v6.0/reference/method/db.collection.find/#mongodb-method-db.collection.find

### Q82. When using the mongoimport command, how can you drop the database before importing?

- Drop the database manually before importing.

Argument: There is no -d option in the [docs](https://www.mongodb.com/docs/database-tools/mongoimport/#options)https://www.mongodb.com/docs/database-tools/mongoimport/#options.

Note: Assuming you are asked to drop a collection instead while importing, the use the --drop option.

arpitajana1220:
## PR Checklist

### DOD

- [  ] I have added new quiz{'s}
- [ x] I have added new reference link{'s}
- [ x] I have made small corrections/improvements

### Changes

1. _Corrected answers_
2. _added reference_
3. _added Note_

### DIFF
--- a/mongodb/mongodb-quiz.md
+++ b/mongodb/mongodb-quiz.md
@@ -165,8 +165,10 @@
 
 - [ ] Calculate interest quickly.
 - [ ] Accomplish nothing, since compound indexes aren't allowed in Mongo.
-- [ ] Use more than one field per index.
-- [x] Combine fields in different collations.
+- [x] Use more than one field per index.
+- [] Combine fields in different collations.
+
+[Reference](https://www.mongodb.com/docs/v6.0/core/index-compound/)
 
 #### Q25. Why are ad-hoc queries useful?
 
@@ -314,8 +316,10 @@
 
 - [ ] restore
 - [ ] read/write
-- [x] dbadmin
-- [ ] delete collections
+- [] dbadmin
+- [x] delete collections
+
+[Reference](https://www.mongodb.com/docs/v6.0/reference/built-in-roles/#backup-and-restoration-roles)
 
 #### Q46. Which MongoDB shell command deletes a single document?
 
@@ -323,6 +327,10 @@
 - [ ] `db.customers.drop({_id: 1});`
 - [ ] `db.drop.customers({_id: 1});`
 - [x] `db.customers.remove({_id: 1});`
+
+__Note:__ db.collection.remove() is deprecated in the new mongosh. Use db.collection.deleteOne() or db.collection.deleteMany().
+
+[References](https://www.mongodb.com/docs/mongodb-shell/crud/delete/)(https://www.mongodb.com/docs/v6.0/reference/method/db.collection.remove/)
 
 #### Q47. Using the MongoDB shell, how do you remove the customer collection and its indexes?
 
@@ -345,12 +353,20 @@
 - [ ] `db.people.getIndexKeys();`
 - [x] `db.people.getIndexes();`
 
+__Note:__ An alternative method in the mongosh shell is listIndexes()
+
+[Reference] (https://www.mongodb.com/docs/v6.0/reference/command/listIndexes/#mongodb-dbcommand-dbcmd.listIndexes)
+
 #### Q50. You are going to do a series of updates to multiple records. You find setting the multi option of the update() command too tiresome. What should you do instead?
 
 - [ ] Use the replaceMany() command instead
 - [ ] Use the updateMulti() command instead
 - [x] Use the updateMany() command instead
 - [ ] Set the global multi option to True
+
+__Note:__ An alternative method for db is .update()
+
+[Reference:] (https://www.mongodb.com/docs/v6.0/reference/command/update/#mongodb-dbcommand-dbcmd.update)
 
 #### Q51. To cleanly shut down MongoDB, what command should you use from the MongoDB shell?
 
@@ -365,6 +381,10 @@
 - [ ] `db.members.find({$match: {gender: "Female"}}, {$group: {_id: {city: "$city"}, number: {$sum: 1}}}.$sort ({number: -1})`
 - [ ] `db.members.find([ {$match: {gender: "Female"}}, {$group: {_id: {city: "$city"}, number: {$sum: 1}}}, {$sort :{number: -1}}])`
 - [ ] `db.members.aggregate([ {$match: {gender: "Female"}}, {$sort :{number: -1}}])`
+
+__Note:__ If you want to analyze the performance of a query use .explain("executionStats")
+
+[Reference:] (https://www.mongodb.com/docs/v6.0/tutorial/analyze-query-plan/)
 
 #### Q53. When no parameters are passed to `explain()`, what mode does it run in?
 
@@ -383,7 +403,7 @@
 #### Q55. Which file in the MongoDB directly holds the MongoDB daemon?
 
 - [ ] mongodb
-- [x] mongo-daemon
+- [ ] mongo-daemon
 - [ ] daemon
 - [x] mongod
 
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Hello, @arpitajana1220
Thank you for your great contribution! 👍 
Have a great day!

Ebazhanov:
@all-contributors please add @arpitajana1220 for design, code

PR #6081, opened 2023-09-21
Update bash-quiz.md

jmarquez-cs:
The script you provided enables extended globbing with `shopt -s extglob` and then attempts to trim the whitespace from the beginning and end of the VAR variable using parameter expansion combined with the extended globbing patterns.

Let's break down the commands:

1. `VAR=${VAR##+([[:space:]])}`:
This command trims the leading whitespaces.
  `##` trims from the start of the string.
  `+([[:space:]])` matches one or more whitespace characters.   
2. `VAR=${VAR%%+([[:space:]])}`:
This command trims the trailing whitespaces.
  `%%` trims from the end of the string.
  `+([[:space:]])` matches one or more whitespace characters.   
  
Given the initial value of VAR as ' This is... a string of characters ', after the trimming operations, the value would be:

`This is... a string of characters`

So, the correct output would be:

`<pre>This is...     a string of characters</pre>`

## PR Checklist

### DOD

### Changes / Instructions

_Add instructions to me, please type here, thanks_

### DIFF
--- a/bash/bash-quiz.md
+++ b/bash/bash-quiz.md
@@ -1020,7 +1020,7 @@ echo "$VAR"
 
 - [ ] `<pre> This is... a string of characters</pre>`
 - [ ] `<pre> This is...a string of characters</pre>`
-- [ ] `<pre>This is... a string of characters</pre>`
+- [x] `<pre>This is... a string of characters</pre>`
 - [ ] `<pre>This is...a string of characters</pre>`
 
 References:


Ebazhanov (approved):
Thanks for the answer, looks great! @jmarquez-cs 👍

Ebazhanov:
@all-contributors please add @jmarquez-cs for design, code

PR #6197, opened 2023-10-05
Corrected answer of json Q17

just-ctrlC-ctrlV:
## PR Checklist

### DOD

- [X] I have made small correction/improvements

### Changes / Instructions

_Add instructions to me, please type here, thanks_

### DIFF
--- a/json/json-quiz.md
+++ b/json/json-quiz.md
@@ -237,8 +237,8 @@
     5.Requests information from a specific source.
     6.Removes information.
 
-- [x] 1.POST 2.DELETE 3.OPTIONS 4.HEADER 5.GET 6.PUT
-- [ ] 1.POST 2.PUT 3.HEADER 4.OPTIONS 5.GET 6.DELETE
+- [] 1.POST 2.DELETE 3.OPTIONS 4.HEADER 5.GET 6.PUT
+- [x] 1.POST 2.PUT 3.HEADER 4.OPTIONS 5.GET 6.DELETE
 
 #### Q18. Which code uses valid JSON syntax for the text "Centennial Olympics?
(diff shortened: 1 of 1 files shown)


just-ctrlC-ctrlV:
Here is a more detailed explanation of each method:

POST: The POST method is used to create or update information on a server. It can be used to submit a form, upload a file, or create a new resource. The POST method allows the client to send data to the server in the body of the request.
PUT: The PUT method is used to update an existing resource on a server. It is idempotent, meaning that it can be called multiple times without causing any unwanted side effects. The PUT method allows the client to send the updated resource to the server in the body of the request.
HEAD: The HEAD method is used to retrieve the headers of a resource without retrieving the body of the resource. This can be useful for checking the existence of a resource or for getting information about the resource, such as the last modified date or the content type.
OPTIONS: The OPTIONS method is used to determine the communication options available for a resource. This can be useful for finding out what HTTP methods are supported by a resource or for getting information about the authentication mechanisms required to access a resource.
GET: The GET method is used to retrieve a representation of a resource from a server. The GET method is idempotent and safe, meaning that it can be called multiple times without causing any unwanted side effects. The GET method does not allow the client to send data to the server in the body of the request.
DELETE: The DELETE method is used to delete a resource from a server. The DELETE method is not idempotent, meaning that calling it multiple times may have different results. The DELETE method does not allow the client to send data to the server in the body of the request.

Ebazhanov (approved):
Thank you @just-ctrlC-ctrlV for correction, LGTM 👍

Ebazhanov:
@all-contributors please add @just-ctrlC-ctrlV for design, code

PR #6542, opened 2023-10-17
Changed Q11 answer

castilhera:
## PR Checklist

### DOD

### Changes / Instructions

An lvalue does indeed represent an object or a memory location, and objects generally have addresses in memory. However, an lvalue is not solely defined by whether it represents an object with an address. It also involves the ability to assign a value to that object or location, typically through a pointer or reference.

So, while an lvalue often represents an object with an address, it's more accurately described as an expression that represents a memory location suitable for assignment, which typically includes variables, array elements, and dereferenced pointers or references.

That said, I changed the answer from:
> It's an expression that represents an object with an address.

To:
> It's a location value, meaning a memory address suitable for assigning to a pointer or reference.

### DIFF
--- a/c++/c++-quiz.md
+++ b/c++/c++-quiz.md
@@ -170,9 +170,9 @@ typedef struct {
 #### Q11. What is an lvalue?
 
 - [ ] It's a constant expression, meaning an expression composed of constants and operations.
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Looks good! 👍

PR #6564, opened 2023-10-18
Update jquery Q72 answer

castilhera:
## PR Checklist

### DOD

### Changes / Instructions

The statement "None of these - jQuery requires AJAX" is not correct. You can use jQuery for various purposes and it does not require AJAX or animation in all projects.

Although this question is confusing, maybe a trick question, because the answers "jQuery 2", "jQuery 3 compressed", and "jQuery 3 slim" are correct, "jQuery 3 slim" is the best option.

It is recommended over jQuery 2 because it includes bug fixes, performance improvements, and support for modern browsers that jQuery 2 may not have and is a smaller, lighter version of jQuery 3 that excludes the AJAX and animation modules. This makes it suitable for projects that primarily require DOM manipulation without needing the additional features provided by the full version of jQuery.

### DIFF
--- a/jquery/jquery-quiz.md
+++ b/jquery/jquery-quiz.md
@@ -1424,9 +1424,9 @@ $('#element')
 #### Q72. If your JavaScript project involves a lot of DOM manipulation, but no AJAX or animation, which version of jQuery should you use?
 
 - [ ] jQuery 3 compressed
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
👍

PR #6966, opened 2023-12-12
suggestion: add explanation to answer of HTML quiz Q55

oorkan:
## PR Checklist

### DOD

### Changes / Instructions

The answer to HTML quiz question 55 is correct but may be confusing because the `<br>` element can be useful for writing a poem or an address (See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br)). The whole matter of not using it and instead using paragraphs with CSS styling is the accessibility ([see MDN -> `<br>` -> Accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility_concerns)).

I suggest adding a little explanation so the reader knows why the chosen option is the correct one.

Thank you!


oorkan:
Thank you, @Ebazhanov !

PR #6967, opened 2023-12-12
fix: answer of HTML quiz Q106 is incorrect

oorkan:
## PR Checklist

### DOD

### Changes / Instructions

The answer to HTML quiz question 106 is incorrect.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th):

```
The <th> HTML element defines a cell as the header of a group of table cells. The exact nature of this group is defined by the scope and headers attributes.
```

and 

image

Plus, when we compare the options **A** and **B**, here's what we get in the browser:

**option A** (❌ **incorrect**)

image

**options B** (✅ **correct**)

image

Summing up, the correct answer is option **B**:

```html
<table>
  <tr>
    <th scope="row">Header</th>
    <td>10</td>
    <td>18</td>
  </tr>
</table>
```

Thank you!

### DIFF
--- a/html/html-quiz.md
+++ b/html/html-quiz.md
@@ -1784,7 +1784,7 @@ As Steve Krug once said, happy talk must die.
 
 #### Q106. How would you mark up a header for a table row?
 
-- [x] A
+- [ ] A
 
 ```HTML
 <table>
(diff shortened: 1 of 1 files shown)


Ebazhanov (approved):
Great job, thank you @oorkan! 👍

PR #7003, opened 2024-02-15
Update, Correct, Replace, and Add questions Ruby on Rails Assessment

RickHPotter:
## PR Checklist

### DOD

### Changes / Instructions

I have added comments for instruction as well.

### DIFF
--- a/ruby-on-rails/ruby-on-rails-quiz.md
+++ b/ruby-on-rails/ruby-on-rails-quiz.md
@@ -214,7 +214,7 @@ class ProductController < ActionController::Base
 
   def update
     @product = Product.find(params[:id])
-    if @product.update_attributes(product_params)
+    if @product.update(product_params)
       redirect_to(product_path(@product))
     else
       render('edit')
@@ -233,6 +233,8 @@ end
 - [ ] The product will not be updated and the controller will raise an ActiveModel::ForbiddenAttributes exception.
 - [x] The product will be updated with the values for name, style, and color, but the value for price will be ignored.
 - [ ] The product will be updated with the values for name, style, color, and price.
+
+[Reference](https://stackoverflow.com/a/19277971)
 
 #### Q25. A Rails project has ActiveRecord classes defined for Classroom and Student. If instances of these classes are related so that students are assigned the ID of one particular classroom, which choice shows the correct associations to define?
 
@@ -337,11 +339,11 @@ end
 - [ ] The method will ignore exceptions that occur during execution.
 - [x] It is a more powerful or destructive version of the method.
 
-#### Q33. What part of the code below causes the method `#decrypt_data` to be run?
+#### Q33. What options below would cause the method `#decrypt_data` to be run?
 
 ```rb
 class MyModel < ApplicationRecord
-after_find :decrypt_data
+  after_find :decrypt_data
 end
 ```
 
@@ -349,6 +351,8 @@ end
 - [ ] `MyModel.where(id: 42)`
 - [x] `MyModel.first.destroy`
 - [ ] `MyModel.new(field: 'new instance')`
+
+[Reference](https://guides.rubyonrails.org/active_record_callbacks.html#running-callbacks)
 
 #### Q34. Which Rails helper would you use in the application view to protect against CSRF (Cross-Site Request Forgery) attacks?
 
@@ -472,6 +476,8 @@ end
 <input type="checkbox" name="post[visible]" value="1" />
 ```
 
+[Reference](https://apidock.com/rails/ActionView/Helpers/FormHelper/check_box)
+
 #### Q40. There is a bug in this code. The logout message is not appearing on the login template. What is the cause?
 
 ```ruby
@@ -496,6 +502,8 @@ end
 - [ ] All attributes in an ActiveRecord model are read-only declared as writable using attr_accessible
 - [x] An instance of an ActiveRecord model will have attributes that match the columns in a corresponding database table.
 - [ ] ActiveRecord models can have only attributes that have a matching database column
+
+[Reference](https://api.rubyonrails.org/classes/ActiveRecord/Base.html)
 
 #### Q42. What is the correct way to assign a value to the session?
 
@@ -522,6 +530,8 @@ session_save(:user_id, user.id)
 ```ruby
 session[:user_id] = user.id
 ```
+
+[Reference](https://guides.rubyonrails.org/v7.1/action_controller_overview.html#session)
 
 #### Q43. Which choice best describes the expected value of @result?
 
@@ -594,7 +604,6 @@ end
 
 - [ ] D
 
-```ruby
 ```ruby
 class Project
   validates_presense_of :name, :unique => true
@@ -691,6 +700,8 @@ saved_name = _____
 - [ ] category.changes[:name]
 - [ ] category.name_changed?
 
+[Reference](https://api.rubyonrails.org/v7.1/classes/ActiveModel/Dirty.html#method-i-2A_was)
+
 #### Q53. Given two models, what is the issue with the query used to fetch them?
 
 ```ruby
@@ -756,6 +767,8 @@ after_update_commit do
 Rails.cache.destroy(cache_key)
 end
 ```
+
+[Reference](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-delete)
 
 #### Q57. After this migration has been executed, which statement would be true?
 
@@ -851,6 +864,7 @@ end
 - [ ] `@user.city`
 - [ ] `@user.try(:city)`
 
+[Reference](https://apidock.com/rails/Module/delegate)
 [Reference](https://itnext.io/understanding-delegate-in-ruby-on-rails-i-wish-i-knew-before-5edd341bad47)
 
 #### Q62. Where would this code most likely be found in a Rails project?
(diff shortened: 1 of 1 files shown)


ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -230,8 +230,8 @@ end
 ```
 
 - [ ] The product will not be updated and the edit template will be rendered.
-- [x] The product will not be updated and the controller will raise an ActiveModel::ForbiddenAttributes exception.
-- [ ] The product will be updated with the values for name, style, and color, but the value for price will be ignored.
+- [ ] The product will not be updated and the controller will raise an ActiveModel::ForbiddenAttributes exception.
+- [x] The product will be updated with the values for name, style, and color, but the value for price will be ignored.
```
RickHPotter, line 234: Before in Rails 4, the B was the answer because it defaulted to always raising an error when unpermitted parameters were attempted to used to create/update a record.

However the default today is to only log, and this can be changed using `config.action_controller.action_on_unpermitted_parameters = :raise` as can be seen [here](https://guides.rubyonrails.org/configuring.html#config-action-controller-action-on-unpermitted-parameters).

RickHPotter: I feel attacked because the [rails guide](https://guides.rubyonrails.org/getting_started.html#using-strong-parameters) says the exception is thrown, but when I test this, this is not what happens.

Even if I use the shebang save! in my create controller, only this log occurs without any exception being raised, (the data is saved as long as the unpermitted param is not required, a ActiveRecord::RecordInvalid is raised in case I use save! instead of save, but that's not related).

image

RickHPotter: I decided to make the change from `update_attributes` to `update` as this is better for both old as new versions of Rails. And I decided to correct the answer based on testing (in real life, I can only remember trying to debug why a piece of data is not being saved to database exactly due to this exception not being raised).

I'm down with keeping the original answer as well, given that apart from the test I did, I couldn't come up with some solid source backing me up.

RickHPotter: I could reproduce what the guides say about raising the exception, but there's a problem. in a controller, that's not the default behaviour used.

image

In the controller, first of all, the params itself is a hash that contains some other data apart from the submitted form.

```ruby
11:04:41 web.1  |   Parameters: {"authenticity_token"=>"[FILTERED]", "card_transaction"=>{"user_id"=>"1", "ct_description"=>"123", "ct_comment"=>"1231", "user_card_id"=>"5", "price"=>"111", "date"=>"2024-02-15"}, "commit"=>"Submit"}
```

As you can see, what we actually get is params[:card_transaction], for example. Inside that, we get the attributes. And that is what `params.require(:card_transaction)` does.

My test in the console was as follows:

image

The reason for this is simple: the `user_params` we use for `User.new` is the result of `params.require(:user).permite(:email)` which **excludes** the entries that are not permitted in the original `params` and therefore, the error is not caught [here](https://github.com/rails/rails/blob/ed76d0fbc74545f0275506b4ebf2dfd375fdda18/activemodel/lib/active_model/forbidden_attributes_protection.rb#L23).

RickHPotter: I'm very sorry for making this so big. [😔](https://emojipedia.org/pensive-face)

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -345,9 +345,9 @@ after_find :decrypt_data
 end
 ```
 
-- [ ] `MyModel.first.update(field: 'example')`
-- [x] `MyModel.where(id: 42)`
-- [ ] `MyModel.first.destroy`
+- [x] `MyModel.first.update(field: 'example')`
```
RickHPotter, line 348: It works even when destroying, because the important part is the method that comes before destroying.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -685,9 +686,9 @@ category.name = 'News'
 saved_name = _____
 ```
 
-- [ ] category.name_was
+- [x] category.name_was
 - [ ] category.saved(:name)
-- [x] category.changes[:name]
+- [ ] category.changes[:name]
```
RickHPotter, line 691: I went as back as Rails 3.0.9 documentation and could not find a place where `model_instance.changes[attribute.to_sym]` would ever return a String instead of a hash containing attributes as keys and pairs as [old_value, new_value].

```ruby
person.changes # => {}
person.name = 'bob'
person.changes # => { 'name' => ['bill', 'bob'] }
```

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -795,13 +796,13 @@ rescue => e
...
 
-- [x] B
+- [ ] B
 
 ```ruby
 rescue_from ActiveRecord::RecordNotFound, with: :render_not_found_response
 ```
 
-- [ ] C
+- [x] C
```
RickHPotter, line 805: Apart from the Reference added that states the usage of cache.delete, I have not been able to find a single mention of `cache.destroy`.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -829,7 +830,7 @@ end
 - [ ] `<%= render partial: "nav", globals: {selected: "about"} %>`
 - [ ] `<%= render partial: "nav", local_variables: {selected: "about"} %>`
 - [x] `<%= render partial: "nav", locals: {selected: "about"} %>`
-- [ ] `<%= render partial: "nav", selected: "about"} %>`
+- [x] `<%= render partial: "nav", selected: "about"} %>`
```
RickHPotter, line 833: This is pretty straightforward. You don't need locals anymore, in fact, you don't need partial keyword anymore either.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -846,8 +847,8 @@ end
 ```
 
 - [ ] `@user.user_city`
-- [ ] `@user.location_city`
-- [x] `@user.city`
+- [x] `@user.location_city`
+- [ ] `@user.city`
```
RickHPotter, line 851: I was unable to find a documentated rails version that implied that `@user.city` would ever be the right answer. Unless there was a point in which `prefix: true` didn't exist, `@user.location_city` is, in all cases, the right answer.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -1017,9 +1018,9 @@ end
 
 #### Q76. Once this form is submitted, which code in the controller would retrieve the string for :name?
```
RickHPotter, line 1019: [form_for](https://apidock.com/rails/ActionView/Helpers/FormHelper/form_for) works different than [form_with](https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormHelper/form_with).

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -897,18 +912,12 @@ end
 - [ ] It includes the number of associated images when determining how many records to return.
 - [x] It will execute two database queries of 21 database queries.
 
-#### Q67. What line of code causes the method `decrypt_data` to be run?
```
RickHPotter, line 900: This question is duplicate of Q33, therefore I replaced it with a new question.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -969,12 +978,37 @@ end
 
 [Reference](https://guides.rubyonrails.org/layouts_and_rendering.html)
 
-#### Q71. Which choice would you not consider when selecting a gem for your project?
```
RickHPotter, line 972: This question seems too opinionated, therefore I replaced it with a new question.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -1011,15 +1047,15 @@ end
 `<%= render partial: 'user_info', object: { name: 'user' } %>`
 
 - [ ] `<%= locals.user_info.name %>`
-- [x] `<%= object.name %>`
+- [ ] `<%= object.name %>`
```
RickHPotter, line 1050: Hash, in Ruby or Rails, does not accept dot notation, an OpenStruct does, but I checked the type of locals and it's just a Hash. I also tested this, and `object.name` returns an error.

Because of this, I decided to create a new (correct) option.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -1059,4 +1097,53 @@ end
 - [ ] `To seed the database with initial data.`
 - [ ] `To test the database connection.`
 
-[reference](https://guides.rubyonrails.org/v3.2/migrations.html#:~:text=Active%20Record%20tracks%20which%20migrations,the%20structure%20of%20your%20database.)
+[Reference](https://guides.rubyonrails.org/v3.2/migrations.html#:~:text=Active%20Record%20tracks%20which%20migrations,the%20structure%20of%20your%20database.)
+
+#### Q79. What is the execution result of the following Ruby code?
```
RickHPotter, line 1102: New question a little bit more ruby-related concerning how self (although consistant) differs inside class / instance methods in a Class.

ruby-on-rails/ruby-on-rails-quiz.md
```diff
@@ -1059,4 +1097,53 @@ end
...
+```
+
+- [ ] `Class Class`
+- [ ] `A A`
+- [ ] `A Class`
+- [x] `Class A`
+
+[Reference](https://www.ignacio.al/self-class-ruby.html)
+
+#### Q80. Given the following model class and view , which code snippet should you use to fetch the correct results, and avoid the N+1 query issue?
```
RickHPotter, line 1125: New question concerning rails N+1 queries. It's not the first one, but it's one I haven't seen everywhere.

Ebazhanov (approved):
Hello, @RickHPotter! 

Thank you for your great job and comments ❤️ !

Ebazhanov:
@all-contributors please add @RickHPotter for design, code

PR #7104, opened 2025-10-15
Fix incorrect answer for Kotlin Q1 and update explanation (Closes #7096)

### LINKED ISSUE #7096: Kotlin Q1's answer is incorrect
Q1. You would like to print each score on its own line with its cardinal position. Without using var or val, which method allows iteration with both the value and its position?
```
fun main() {
  val highScores = listOf(4000, 2000, 10200, 12000, 9030)
}
```
Actual Answer

- [x]  .withIndex()
- [ ]  .forEachIndexed()
- [ ]  .forEach()
- [ ]  .forIndexes()

Expected Answer

- [ ]  .withIndex()
- [x]  .forEachIndexed()
- [ ]  .forEach()
- [ ]  .forIndexes()

Explanation:
`withIndex()` returns an `Sequence<IndexedValue<T>>` which still holds a Sequence but does not iterate through it. The correct answer should be `.forEachIndexed()` which does allow iteration through the Sequence with both the value and its position.

DheerajSChauhan:
# Fix incorrect answer for Kotlin Q1 - forEachIndexed vs withIndex

## Description
This PR fixes the incorrect answer for Kotlin Question 1 about iterating with both value and position without using `var` or `val`.

## Issue
Closes #7096

## Current Problem
- **Marked as correct:** `.withIndex()`
- **Actually correct:** `.forEachIndexed()`

## Changes Made
Updated the correct answer from `.withIndex()` to `.forEachIndexed()` in the Kotlin quiz.

### Why `.withIndex()` is INCORRECT:
1. **Does not iterate by itself** - It only returns an `Iterable<IndexedValue<T>>`
2. **Requires additional methods** - You must chain it with `.forEach()` or similar
3. **Implicit variable usage** - Destructuring `IndexedValue` creates variables internally
4. **Doesn't meet the constraint** - The question specifically asks for a method "without using var or val"

Example of `.withIndex()` (requires additional steps):
```kotlin
highScores.withIndex().forEach { (index, score) ->
    println("${index + 1}. $score")
}
```

### Why `.forEachIndexed()` is CORRECT:
1. **Directly iterates** through the collection
2. **Provides both index and value** as lambda parameters
3. **No var/val needed** - Parameters are passed directly to the lambda
4. **Single method call** - Does exactly what the question asks

Example of `.forEachIndexed()` (correct approach):
```kotlin
highScores.forEachIndexed { index, score ->
    println("${index + 1}. $score")
}
```

## Testing
- Verified the code example works correctly in Kotlin
- Confirmed `.forEachIndexed()` meets all requirements stated in the question
- Checked Kotlin official documentation for accurate behavior

## File Changes
- [ ] `kotlin/kotlin-quiz.md` - Updated correct answer marker
- [ ] `kotlin/kotlin-quiz.md` - Updated explanation (if applicable)

## Additional Context
The question explicitly asks for a method that allows iteration with both value and position **without using var or val**. Only `.forEachIndexed()` satisfies all these requirements in a single method call.

---

**References:**
- [Kotlin Documentation - forEachIndexed](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each-indexed.html)
- [Kotlin Documentation - withIndex](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/with-index.html)

### DIFF
--- a/kotlin/kotlin-quiz.md
+++ b/kotlin/kotlin-quiz.md
@@ -8,12 +8,12 @@ fun main() {
 }
 ```
 
-- [x] `.withIndex()`
-- [ ] `.forEachIndexed()`
+- [ ] `.withIndex()`
+- [x] `.forEachIndexed()`
 - [ ] `.forEach()`
 - [ ] `.forIndexes()`
 
-[reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/with-index.html)
+[reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/for-each-indexed.html)
 
 #### Q2. When the **Airplane** class is instantiated, it displays **Aircraft = null**, not **Aircraft = C130** why?
 


Ebazhanov:
@all-contributors please add @DheerajSChauhan for design, code

PR #7249, opened 2025-11-25
Add OOP questions Q19, Q51, Q53, Q57, Q60, Q63, Q67, Q69, Q88, Q90, Q92, Q93, Q103, Q105, Q106 (#5797)

### LINKED ISSUE #5797: Set of questions for OOP that got me into the top 30% of applicants
### Q57. What is a copy constructor?
 It is a unique constructor for creating a new object as a copy of an object that already exists. There will always be only one copy constructor that can be either defined by the user or the system.

### Q69. What is the most accurate example of the Liskov substitution principle?
```
public class Car{
}
public class FlyingCars extends Car{
   public void fly(){}
}
public class Tesla extends FlyingCar{}
public class Honda extends Car{}
```

### Q105. object are passed by ___?
 value or reference, depending on the programmin language used

### Q53. What is the function of a user diagram?
 It links actors to roles played in all use cases.

### Q60. You have modules that are dependent on each other. If you change one module, you have to make changes in the dependent modules. What term is used to describe this problem, and what is a potential solution?
 Coupling. A solution is to refactor the code to be loosely coupled by using inversion of control and dependency injection.

### Q88. In addition to responsibilities, what should be listed on Class-responsability-collaboration (CRC) cards?
 attributes

### Q19. Which statement is true?
 A default parameter's constructor is not equivalent to the default constructor

### Q90. Which relationship best illustrates an abstract-concrete class relationship?
 planet : moon

### Q106. What is a method?
 a function associated with data and behavior

### Q51. What is a reference to an object?
 It is the address where the variables and methods of an object are stored.

### Q67. Which statements best describe the Gang of Four design patterns called Memento and Observer?
 Memento captures and restores an object's internal state. Observer notifies multiple classes of changes.

### Q103. If a language just implements pass-by-value parameters to functions, can the objects passed as parameters be set to new instances inside the function and their new value will return to the caller?
 No, a local copy will be initialized and will be lost as soon as it is out of scope.

### Q92. How many levels does multilevel inheritance allow in a program?
 only the amount of levels memory permits, divided by processor speed

### Q93. What is a virtual Method?
 a method that you expect may be redefined in derived classes

### Q63. Which choice is a benefit of using dependency injection?
 loose coupling

KL2400040448:
Fixes #5797

Added multiple OOP interview questions with correct answers and formatting:
Q19, Q51, Q53, Q57, Q60, Q63, Q67, Q69, Q88, Q90, Q92, Q93, Q103, Q105, Q106.

These align with the repository's quiz style and improve OOP coverage.
Screenshot 2025-11-25 212414
Screenshot 2025-11-25 211616


Ebazhanov (changes requested):
@KL2400040448 pls doulbe check PR has no changes

Ebazhanov (changes requested):
still empty

